Issue #2, opened 2016-07-08
add the  bitshuffle implementation used for new eiger detectors


kif:

vasole:
It should be accomplish by #5.

kif:
close #5

PR #5, opened 2016-07-09
Add bitshuffle versions for VS2015 (Python 3.5) and VS2008 (Python 2.7)

vasole:
@kif 

Addresses #2 for Python 3.5

I have compiled the libraries as Release with Debug information. They are bigger but they are supposed to be optimized and have a chance of providing information in case of crash.


vasole:
closes #2

PR #18, opened 2019-10-02
Rework hdf5plugin to add compression and blosc filter support

t20100:
This PR proposes a major rework of `hdf5plugin`.
It :

- Provides decompression AND compression by using a dynamic loader on Linux and macOS to access HDF5 function without linking to the HDF5 library, and by building the the filters as dll on Windows (where linking is handled differently).
- Adds the blosc filter.
- Builds filters from source by embedding filter sources with `git subtree`
- Include hdf5 headers for simpler build. This might not be the best idea because some headers like `H5pubconf.h` are generated during the build.... but if it works it simplifies build and CI. 

It changes the behavior of `hdf5plugin` in mainy ways including:
- It not more set the `HDF5_PLUGIN_PATH` environment variable
- It provides compression
- It is meant to be used with `h5py` and registers the filters in the HDF5 library used by `h5py`. That being said, one should be able to use the filters as HDF5 plugins for decompression as the dynamic loading wrapper provides no-op fallbacks if the dynamic loading is not initialized.

This needs more testing, but it should work on Windows, Linux and macOS with HDF5 1.8 and 1.10.


t20100:
git subtree of https://github.com/nexusformat/HDF5-External-Filter-Plugins is overkill... as we use one file from there.... I propose to copy this one file instead of using `git subtree`.

vasole:
OK, let's copy the file.

t20100:
I removed the https://github.com/nexusformat/HDF5-External-Filter-Plugins subtree, and make setup.py python 2.7 friendly...

t20100:
Fixed macOS CI environment.
Added `H5pubconf.h` built with VS2008-64bits.

More testing is needed: with HDF5-1.8, 32bits OS, usage as plugins for other apps, ...
Including the `H5pubconf.h`  header is probably the weak point of it.

t20100:
- Added compile flags for bitshuffle (issue on centos7)
- Added CI also testing with oldest available wheels of h5py (so with HDF5 1.8.x except for python3.7)

vasole:
- The wheels built on Centos7 work on Debian 9
- The generated shared libraries work with MatLab as plugins in Debian 9

vasole:
I think the test suite should be shipped with the wheel.

I am not fully satisfied with the names of the generated libraries. I would rename it to skip the python specific naming.

t20100:
I'll take care of the last 2 comments

PR #27, opened 2019-10-04
Add miniconda37 to test env on windows

t20100:
This PR:
- Adds test with miniconda3
- Remove use of virtualenv (not really needed)


kif:
LGTM

PR #34, opened 2019-10-07
Fix typo and clean-up

t20100:
This PR:
- Fixes a typo in the blosc compression name to id.
- Adds a `_` prefix to everything that is not public in the `__init__.py` file.


t20100:
Updated license info (as there is a bit of mixed licenses).

vasole:
I'm not sure about the need to put multiple licenses in the setup. We'll see.

t20100:
Me neither, but the fact is it's a mix of licenses: MIT, BSD, BSD-2 clauses, zlib.
I don't know how to deal with that.

PR #37, opened 2019-10-08
updated documentation

Florian-toll:
updated the read me


kif:
Apparently you committed some notebooks and HDF5 files. You should maybe restart your branch with only the updated file.

Issue #44, opened 2019-10-18
Timing for 2.0 release?


rlizzo:
Hi team, just wondering what the timing for the v2.0 release will be? I saw the pre-release a few days, ago, and got very excited about being able to include it in [the project I maintain](https://github.com/tensorwerk/hangar-py) in our upcoming 0.4 release (which I'd like to get out in the next two weeks or so).

If there's anything you could use some help on, I'd be happy to contribute a PR or two to help speed along the process. Just let me know!

Best,
Rick

vasole:
We just wanted feedback from the community. I do not think we have any showstopper to make a release.

rlizzo:
Ahh excellent! Thanks for the quick reply!

I've had a remarkable amount of success with the pre-release. Everything seems to go as expected.

A few notes about the documentation though:
- state which versions of python this is compatible with (is python 3.8 supported? I haven't tried it myself yet)
- state that the project is available on linux, macos, and windows (windows support is something that is missing from the `blosc-hdf5-plugin` package on `conda-forge`.  

In addition, it's not explained why it states in the readme 

> "Whenever possible, HDF5 compression filter plugins are best installed system-wide or through Anaconda (blosc-hdf5-plugin, hdf5-lz4)."

I'm actually not sure of that myself? 

Finally, I know there was an issue about this awhile ago in #17, but is there any plan to add this to the `conda-forge` anaconda channel?  I only have to say that this would be beneficial for projects which desire to keep their depedencies on `pypi` and `conda-forge` in sync. I've got some experience packaging for both independent `anaconda` channels as well as `conda-forge`, and would be happy to help put together a `conda-forge feedstock` with the team after 2.0 is finally released!

rlizzo:
Also, just worth saying thank you for all your hard work on this! It's much appreciated by a fellow open-source project maintainer! This is a phenomenal "quality of life" project which will improve my library and users life drastically!

vasole:
Python 3.8 is supported. The only compiled code are the plugins themselves.

Using the plugins as provided by the system maintainers make sure they were compiled using the same settings used to compile the HDF5 library itself. When one compiles HDF5, there is a file named H5pubconf.h to be used when compiling against the library.

We supply the H5pubconf.h file with the sources of hdf5plugin in order to make sure it can be built. 

Let me illustrate our caveats. On Anaconda, in order to be 100% sure one would have to rely on the H5pubconf.h generated when compiling HDF5 on Anaconda.

[...]

PR #46, opened 2019-10-21
Add --hdf5= option to build against specific version of HDF5

t20100:
This PR adds a `--hdf5=<path>` option to `python setup.py build` in order to specify the HDF5 headers to use (and the .lib on Windows).

related to #44


vasole:
I would prefer to use os.path.join instead of string addition

t20100:
Done (+ added moving `.lib` file for Windows).

PR #49, opened 2019-11-05
Improve bdist_wheel override

### LINKED ISSUE #48: wheel content is stored in platlib subfolder
Investigate why `hdf5plugin` is stored in `hdf5plugin-2.0.0.data/platlib/`.
Check whether this is normal or not.
This is due to setting `root_is_pure` in the subclass of `bdist_wheel` that tweaks the wheel generation to make one wheel working for the different versions of python.

Initially raised here: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/h5py/EMLVilY8oHk

t20100:
This PR updates the bdist_wheel command override by not setting `root_is_pure = True`.
This way, the generated wheel is cleaner.

closes #48

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -48,13 +48,14 @@ else:
         """Override bdist_wheel to handle as pure python package"""
 
         def finalize_options(self):
-            bdist_wheel.finalize_options(self)
-            self.root_is_pure = True
-            self.plat_name_supplied = True
             self.plat_name = get_platform()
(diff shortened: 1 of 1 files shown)


vasole:
Will it work when we get python 4.x.x? (line 53)

t20100:
No, as no wheel would.

elistevens:
Thank you for making this change!

PR #51, opened 2019-11-18
Bitshuffle add option to build with OpenMP

### LINKED ISSUE #24: bitshuffle openmp support
As it is, the bitshuffle HDF5 filter is never built with openmp.

t20100:
This PR adds an option to build the bitshuffle plugin with (or without) OpenMP.
It is enabled by default except on macOS.
To use the option, run: `python setup.py build --openmp=[False|True]`

closes  #24

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -18,13 +18,15 @@ The HDF5 plugin sources were obtained from:
 Installation
 ------------
 
-To install, just run::
+To install, run::
 
-     pip install hdf5plugin
+     pip install hdf5plugin [--user]
 
-To install locally, run::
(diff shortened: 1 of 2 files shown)


vasole:
Should we give up OpenMP support under windows python 2.7 instead of trying to fix things?

I would not spend time on it.

t20100:
Me neither! I'm disabling it for python 2.7 under Windows,

vasole:
Funny. I would have written sys.startswith("win") and sys.version_info < (3,) to identify windows and python 2.7

PR #52, opened 2019-11-18
Add options to use SSE2 and AVX2 (for bitshuffle and blosc)

### LINKED ISSUE #23: blosc filter sse2/avx2 support
As it is, sse2/avx is not taken into account while compiling the HDF5 blosc filter.

t20100:
This PR adds `--sse2` and `--avx2` build options to enable/disable SIMD instructions for blosc and bitshuffle.

closes #23

TODOS:

- [x] Test on Windows
- [x] Test on macOS

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,10 +1,14 @@
-include README.rst
-include MANIFEST.in
-include setup.py
-include setup.cfg
 include .travis.yml
 include appveyor.yml
+include CHANGELOG.rst
+include cpuinfo.py
+include LICENSE
+include MANIFEST.in
+include README.rst
+include setup.py
+include version.py
 recursive-include ci *.py *.sh
+recursive-include doc *
 recursive-include hdf5plugin *.py
-recursive-include test *.py *.h5
 recursive-include src *
+recursive-include test *.py *.h5
--- a/README.rst
+++ b/README.rst
@@ -26,7 +26,7 @@ To install from source and recompile the HDF5 plugins, run::
 
      pip install hdf5plugin --no-binary hdf5plugin [--user]
 
-This can acheive better performances by enabling OpenMP for the bitshuffle plugin.
+Installing from source can achieve better performances by enabling SSE2, AVX2 and OpenMP if available.
 
 Documentation
 -------------
--- a/README.rst
+++ b/README.rst
@@ -26,7 +26,7 @@ To install from source and recompile the HDF5 plugins, run::
 
      pip install hdf5plugin --no-binary hdf5plugin [--user]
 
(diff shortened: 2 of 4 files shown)


t20100:
Not as simple as expected... Enabling AVX2 seems to cause some troubles.

t20100:
Ready for review.

The idea is that:
- wheels are built for "generic CPU" with SSE2 support (as it is part of the x86_64 architecture)  and without AVX2.
- When building from source, by default, it is built:
  * For 'native' CPU on Linux (using all the capabilities of the building CPU at the expense of compatibility)
  * With SSE2 and AVX2 if those are available: there is some checks of compile options availability and CPU capabilities.

The `setup.py build` target have 3 options `native`, `sse2` and `avx2` which allow to control the compilation with a finer grain (those are used to generate the wheels, `build --openmp=False --native=False --sse2=True --avx2=False`).

t20100:
BTW, this PR adds `cpuinfo` (https://pypi.org/project/py-cpuinfo/) to `setup_requires` in order to probe for SSE2 and AVX2 CPU capabilities from Python (so not needed when installing wheels).

kif:
I tried to compile avx2 code on a computer lacking it (AVX1 is present: sandy-bridge)
python3 setup.py build --avx2=True --openmp=True
the bitshuffle binary uses ymm register, not the blosc one.

vasole:
I would prefer not to have additional dependencies.

t20100:
I'm fine with an extra pure-python build dependency, especially that now `setuptools` handles that...
but we can add a copy of `cpuinfo` in the root directory (it's a single file), this is how `python-blosc` is handling it... then we'll have to maintain the copy up-to-date.

t20100:
Added local copy of `cpuinfo` to avoid build dependency.
Also update `MANIFEST.in` (related to #55)

vasole:
Everything is fine except that I would aim at native=False, sse2=True and avx=False as defaults for a good compromise between performance and portability.

I think we need at least the opinion of @kif here.

@tschoonj what is the conda policy concerning SSE2 and AVX2?

tschoonj:
The conda-forge default is to compile on Linux with:

```bash
CFLAGS=-march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe
```

The gcc [docs](https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html) appear to suggest that this means that SSE2 would be available, but not AVX2...

I haven't checked for macOS and Windows, but I assume it will be similar.

vasole:
@tschoonj 

Please, use the options native=False sse2=True avx2=False when building the conda packages for next release.

PR #60, opened 2019-12-02
Rework build to build snappy as a static lib first

### LINKED ISSUE #58: Building with C++11 fails on macOS
The `-std=c++11` is not accepted for C files under macOS.

t20100:
This PR refactors the build to make it work on macOS with C++11.
It builds snappy (C++11) as a static library only if C++11 is used.
This needs a bit a tweaking as checking for compile flags can only be done late in the process with setuptools...

closes #58

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -225,9 +225,11 @@ class PluginBuildExt(build_ext):
         - select compile args for MSVC and others
         - Set hdf5 directory
         """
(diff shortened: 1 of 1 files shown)


t20100:
That should now be working OK.
The checks of compiler options is done early (in the build command) and the snappy (C++11) library is removed from the list before doing the actual build.

PR #64, opened 2020-02-01
replace -openmp with -fopenmp

kif:
it changed the meaning of the option !

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,7 +56,12 @@ else:
         """Override bdist_wheel to handle as pure python package"""
 
         def finalize_options(self):
-            self.plat_name = get_platform()
+            try:
+                self.plat_name = get_platform()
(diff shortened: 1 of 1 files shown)


vasole:
- get_platform failing in all the tests irrespective of Python 2 or Python 3
- the tests are passing on travis python 3.6 despite having the same error as in the other platforms

vasole:
@t20100 

Why not to use distutils.util.get_platform?

t20100:
@kif, nice catch!

@vasole,  I'm using the same `get_platform` as `bdist_wheel` is using as it inherits from it. This is to avoid building wheels for each version of python, but there is maybe other ways.

vasole:
Well, something must be different because it got broken when using wheel but not when using distutils.util

PR #66, opened 2020-02-03
[CI] Use python3 on macos

### LINKED ISSUE #65: [CI] We are testing MacOS exclusively under Python 2
Only python3 testing is actually needed

t20100:
Install miniconda on macos to use python3: closes #65


t20100:
Ready to merge.

CI fails on branch, because the branch was created before the `get_platform` fix was merged, but PR tests are OK.

PR #67, opened 2020-02-04
Prepare 2.1.2 release

t20100:
update changelog and bump version to 2.1.2

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,10 @@
+2.1.2
+-----
+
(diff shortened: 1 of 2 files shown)


vasole:
I do not think the change of continuous integration is relevant to the end users and it should be out of the changelog.

vasole:
Thanks for the attention you give to my comments....

kif:
I did the merge, considering hdf5plugin is not intended to be used by end-users but solely by developers, for whom the indication of CI is of interest (if not, why do most project put CI-badges on their front page ?)

PR #68, opened 2020-04-23
Add support of FCIDECOMP filter plugin

vasole:
This a JPEG-LS filter plugin. It is a lossless codec used in some (astronomy?) applications.


vasole:
@t20100 

This filters needs C++11. Therefore it is incompatible with VS2008 and with the official version of Python 2.7.

t20100:
For C++11, there is already the issue with the snappy compression algorithm in blosc, and there is a flag to toggle use of C++11 (for building manylinux wheels).

t20100:
216 files changed! Impressive!

vasole:
Should I disable then the build of the filter under windows with Python 2.7?

The actual files used is extremely small, but they come with a tar.gz file inside another tat.gz file with the LICENSE and the doc. It is the equivalent of cloning a repository.

vasole:
For the JPEG-LS library I took the latest commit of the master-1.x branch to support Visual Studio 2019 and to avoid the need of C++17 of the master branch. The master-1.x branch is maintained.

In the contributions instructions I added the potential need to add functions to hdf_dl.c

I did not write a helper because I do not know the use of the compression options.

t20100:
For JPEG-LS, since it is not a release, you can put the hash of the commit you used in the doc so it is documented.

t20100:
For C++11 management, I'm looking at what I did for blosc.

t20100:
We can add an almost empty helper just providing the filter ID. For h5py v3, they added support for such class.
You can also add links to the licenses at the bottom of the readme file as for the other filters.

For C++11, we'll have to filter out the extensions when it is not available as it is done for the libraries. I'm trying to make something.

vasole:
I thought the links were there (they had wrong target).

t20100:
CI broken because, fcidecomp tests should only run when the plugin has been built.

t20100:
I added skipping of fcidecomp tests if it is not present (when no c++11).
If it is OK for you and CI is happy, I'll merge it.

vasole:
Fine with me.

Issue #73, opened 2020-05-07
Support for zfp


hhoeflin:
Love hdf5 plugin. Use it for blossc. What would be great it so have lossy compression available. Is there a possibility of supporting zfp.

https://github.com/LLNL/H5Z-ZFP

Thanks!

vasole:
My zfp branch works under windows, but I cannot make it work under linux without passing the location of the hdf5 library as argument:

https://github.com/vasole/hdf5plugin/tree/zfp

@t20100 Should we envisage to include it into the repository under those conditions? The zfp filter under platforms other than windows would have to be built from source passing the location of the HDF5 library. I guess it would work for conda but we would not be able to include zfp support into our wheels.

vasole:
@t20100 The problems come from a plethora of H5?_XXXXXX_g undefined symbols. Most of them from H5E

t20100:
That's due to the weird understanding of constants in libhdf5.
There is already a few in `hdf5_dl.c`:
https://github.com/silx-kit/hdf5plugin/blob/master/src/hdf5_dl.c#L101
https://github.com/silx-kit/hdf5plugin/blob/master/src/hdf5_dl.c#L146
I'll have a look at it.

t20100:
Closed by PR #74.
Related issue: #75

hhoeflin:
Wow that was fast. Thanks so much!

vasole:
@hhoeflin 

Release available from PyPI.

PR #74, opened 2020-05-12
Provides ZFP lossy compression support

vasole:
Addresses #73


t20100:
Fixed test for CI.
There seems to be some parameters that can be provided to the filter (https://h5z-zfp.readthedocs.io/en/latest/interfaces.html#generic-interface)

vasole:
The tests passed for Mac under python 2.7 with latest h5py but not with h5py 2.6.0.

The test passed for Mac under python 3.7 and 2.6.0 but recompiling from source. The oldest binary wheel available for python 3.7 is 2.8.0.

PR #77, opened 2020-05-12
Add Zfp parameters

t20100:
This PR adds optional parameters to the `Zfp` filter in order to configure the compression mode.
It documents those modes in the README and add tests for it.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -189,6 +189,7 @@ The following compression modes are supported:
   For details, see `zfp fixed-rate mode <https://zfp.readthedocs.io/en/latest/modes.html#fixed-rate-mode>`_.
 
   .. code-block:: python
+
         f.create_dataset('zfp_fixed_rate', data=numpy.random.random(100),
             **hdf5plugin.Zfp(rate=10.0))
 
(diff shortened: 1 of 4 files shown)


t20100:
I hesitated for the `Zfp` API:

I first made helper class methods for each mode e.g., `hdf5plugin.Zfp.rate(10)` and the `__init__` taking a mode and its arguments: `hdf5plugin.Zfp(hdf5plugin.Zfp.RATE, rate=10)`.

But since except for the Expert mode is the only one with more than one parameter, it sounds simpler to just write: `hdf5plugin.Zfp(rate=10)`, but maybe that's counter intuitive.

kif:
This would mean:
* hdf5plugin.Zfp(rate=10)
* hdf5plugin.Zfp(precision=5)
* hdf5plugin.Zfp(accuracy=2)
* hdf5plugin.Zfp(minbits=4, maxbits=64, maxprec=8, minexp=2) # to select the expert mode.

I have the feeling that the meaning of those parameters needs to be explained a bit, as they refer to quantities for 4^n data ? For now I did not test it. I would be interested for storing images of the Jungfrau as they have to be float32 after pedestal compression. Especially that the compression is possible using CUDA.

vasole:
We are referring to the zfp doc because things are indeed hard to understand.

Concerning the expert mode, the best is probably to take a look at the explanation of the accuracy mode where they give the equivalent expert mode:

https://zfp.readthedocs.io/en/latest/modes.html#mode-fixed-accuracy

I'm going to play with the fixed accuracy mode. As I understand it, one could use it to get good background compression while keeping **statistically** lossless signals.

PR #79, opened 2020-05-15
Fix readme syntax

t20100:


vasole:
PyPI refused to accept the wheels because of the problem in the long_description. I noticed that one can specify the long_description_type to be other thing than rst, but rst is the default.

PR #89, opened 2020-09-29
enable SIMD on power9 (requires gcc>=8)

kif:
This speeds up the reading of Eiger2 dataset from 36ms to 19ms (on one core) on our power9 computers.
x86 computers are much faster for this task (~12ms)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -78,6 +78,8 @@ def get_cpu_sse2_avx2():
     :returns: (is SSE2 available, is AVX2 available)
     :rtype: List(bool)
     """
+    if platform.machine() == "ppc64le":
+        return True, False
     try:
         import cpuinfo
     except ImportError as e:
@@ -345,6 +347,8 @@ bithsuffle_dir = 'src/bitshuffle'
 # Set compile args for both MSVC and others, list is stripped at build time
 extra_compile_args = ['-O3', '-ffast-math', '-std=c99', '-fopenmp']
 extra_compile_args += ['/Ox', '/fp:fast', '/openmp']
+if platform.machine() == "ppc64le":
+    extra_compile_args += ["-DNO_WARN_X86_INTRINSICS"]
 extra_link_args = ['-fopenmp', '/openmp']
 
 bithsuffle_plugin = HDF5PluginExtension(
--- a/src/bitshuffle/setup.py
+++ b/src/bitshuffle/setup.py
@@ -5,6 +5,7 @@ from __future__ import absolute_import, division, print_function
 import glob
 import os
 from os import path
+import platform
 from setuptools import setup, Extension
 from setuptools.command.build_ext import build_ext as build_ext_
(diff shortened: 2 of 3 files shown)


vasole:
@t20100 

Should we check the compiler version? We are (supposedly) shipping manylinux2014 wheels for power9...

kif:
Not easy to get the version of the compiler from Python if I remember well ...
I did it for Blosc but the build is driven by Cmake.
The same directive exists for clang, but it appeared in version 9 (IIRC).

vasole:
There are distutils "hacks, but the simplest seems to me to recover the output of` ``gcc --version``

t20100:
On Power9, we are now building manylinux2014 wheels and Ubuntu20.04 packages.
Running `gcc --version` on `manylinux2014_ppc64le` gives: `gcc (GCC) 9.3.1 20200408 (Red Hat 9.3.1-2)`
I launched a complete build on your branch to see the outcome: https://gitlab.esrf.fr/silx/bob/hdf5plugin/-/pipelines/34424

It would be best not to modify files under `src/` since it is subtrees of other projects.
There is no need to change ` src/bitshuffle/setup.py` since it is not used, and change in ` src/bitshuffle/src/bitshuffle_core.c` can be replaced by `"-DUSESSE2"` in the project's `setup.py` as it is already done for `"-DNO_WARN_X86_INTRINSICS"`

kif:
On Wed, 30 Sep 2020 01:03:04 -0700
Thomas VINCENT <notifications@github.com> wrote:

> It would be best not to modify files under `src/` since it is subtrees of other projects.
> There is no need to change ` src/bitshuffle/setup.py` since it is not used, and change in ` src/bitshuffle/src/bitshuffle_core.c` can be replaced by `"-DUSESSE2"` in the project's `setup.py` as it is already done for `"-DNO_WARN_X86_INTRINSICS"`

I am not sure about that ...
I see no problem in submitting a PR to bitshuffle

t20100:
Yes, a PR to bitshuffle would be best.

vasole:
Changing line 361 in your setup.py to  extra_compile_args += ["-DNO_WARN_X86_INTRINSICS", "-DUSESSE2"] and not touching any file under src should be enough to allow the desired behavior without waiting for upstream changes.

t20100:
You can have a look at this branch: https://github.com/t20100/hdf5plugin/tree/pr89
No change to `src/bithsuffle` and a slightly more generic approach to enabline SSE2 support on Power9.

kif:
I tried and gcc emits a warning that USESSE2 is already defined ...

kif:
My PR is available upstream to:
https://github.com/kiyo-masui/bitshuffle/pull/87

t20100:
I get no warning with: https://github.com/t20100/hdf5plugin/tree/pr89

t20100:
Closing as supersedes by  #90

PR #101, opened 2020-12-14
Update c-blosc to v1.20.1

vasole:


vasole:
The compiler used with python 2.7 has issues with stdbool.h 
When using more recent visual studio everything is fine.

- Is it worth to update c-blosc to 1.20.1?
- Is it worth to care about python 2.7?

t20100:
I updated version information in the README.

LGTM, I'll merge once CI pass.

PR #105, opened 2020-12-15
Remove Python2.7 compatibility code

t20100:
To merge after PR #104

This PR cleans-up code by removing python 2.7 compatibility code.


vasole:
Are you sure you want to have OpenMP enabled as default under windows?

t20100:
Unless I'm missing something, OpenMP was already enabled for Python 3 on Windows:

https://github.com/silx-kit/hdf5plugin/blob/master/setup.py#L147-L148

vasole:
I find it weird. I do not think it is the policy we have on silx.

t20100:
Actually it is apparently enabled by default in `silx` as well:

https://github.com/silx-kit/silx/blob/7751a07485c0ce4f3ec880af39c67b89308ecae2/setup.py#L543-L544

though the code is fairly complex:

https://github.com/silx-kit/silx/blob/7751a07485c0ce4f3ec880af39c67b89308ecae2/setup.py#L529-L558

vasole:
Then it must be when creating the wheels. I recall we had issues on some user machines when using OpenMP under windows.

t20100:
Yes, I confirm wheels from both `silx` and `hdf5plugin` are built without OpenMP.

Issue #112, opened 2021-03-07
"Cannot initialize filter fcidecomp: File not found"


texadactyl:
This message started appearing in the last several days.

Envirtonment
O/S: Ubuntu 20.10, up-to-date
hdf5plugin version: 2.3.2
h5py version: 3.2.1
Python 3.8.6
All of my packages are managed under pip 20.1.1.

I just recently observed the subject message which seems to originate from the hdf5plugin. 
```
import h5py
import hdf5plugin
print('Right after importing hdf5plugin')
```
gets this stdout:
```
Cannot initialize filter fcidecomp: File not found
Right after importing hdf5plugin
```
Switch the order of the 2 imports, I see the same thing.

texadactyl:
When I did this:
```
pip uninstall -y h5py hdf5plugin
pip install  h5py==2.10.0 hdf5plugin==2.3.2
```
and then ran the above tiny program, this message does not appear.

texadactyl:
So, it looks like an h5py issue.

I'll leave this open so you'll see it but please feel free to mark it as "not my issue"  (-:

vasole:
Not sure where is the problem. Have you tried other h5py 3.x versions?

texadactyl:
Not yet.  2.10 is not alarming to the scientists that I work with.  (-:

texadactyl:
They were worried about their data getting compromised.  2.10 is a version that was on one of the data pipeline servers so I just tried it out.

t20100:
Hi,

I could not reproduce this with Ubuntu 20.04, Python 3.8.5, pip 20.1.1, hdf5plugin 2.3.2, h5py 3.2.1.

takluyver:
It's not obvious to me how this could be an h5py issue, as the code involved in making that error message doesn't seem to touch h5py - it's just finding a file that should be part of the hdf5plugin package:

https://github.com/silx-kit/hdf5plugin/blob/3df5c33cb99add78ef5367a0738276aac033ad3b/hdf5plugin/__init__.py#L301-L306

t20100:
The `fcidecomp` filter requires C++11, so if C++11 is not available, it is not built and you would get this message.
To avoid this message in this case, we would need to store build options (related issue: #96).

Did you installed `hdf5plugin` from source or from wheels?

We provide 2 Linux wheels for `hdf5plugin`:
- `manylinux1`: C++11 not available and so no `fcidecomp` filter.
[...]

PR #113, opened 2021-03-09
Added `hdf5plugin.config` to retrieve build options at runtime; Fixed `FciDecomp` error message when built without c++11

### LINKED ISSUE #96: Build options at runtime ?
When I import hdf5plugin is there a way to check if it was compiled for the current cpu and whether openmp is enabled?
Related, is there a plan to have these options selected at runtime ?
Thanks!

t20100:
This PR adds a `hdf5plugin.config` namedtuple that allows to retrieve the build options (e.g., openmp, cpp11).
This is achieved by writing a python file with a dict definition containing this information from `setup.py`.
In case the build options change between 2 builds, a `clean` is done first to rebuild everything.
The way to deal with this sounds simple enough to me, but if there is a better way to do so, I'll take it.

This PR also avoid error message of `fcidecomp` filter if not build with C++11 and adds an error message when using the `FciDecomp` class.

closes #96, related to #112, hopefully closing it

### DIFF
--- a/hdf5plugin/__init__.py
+++ b/hdf5plugin/__init__.py
@@ -287,6 +287,13 @@ class FciDecomp(_FilterRefClass):
     """
     filter_id = FCIDECOMP_ID
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if not config.cpp11:
+            _logger.error(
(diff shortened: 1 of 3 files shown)


vasole:
LGTM. Unless you want to add a test to verify the presence of the build options dictionary I will merge it.

t20100:
Indeed, done.

PR #114, opened 2021-03-09
Move `hdf5plugin` to a sub-folder

t20100:
Idea is to avoid having a `hdf5plugin` folder in the root of the repo to avoid import from source and thus remove the runtime check of the version file to simplify a bit `setup.py` and `__init__.py`.

I moved `hdf5plugin` into `src/`, but alternatives are possible to avoid mixing plugin sources and `hdf5plugin`:
- it could also be moved to `lib/hdf5plugin`as in matplotlib or whatever sub-folder name.
- `hdf5plugin` could be renamed to `lib` or anything.

What do you think?


t20100:
attn @vallsv @kif what do you think of it?
Since we're going to make a release soon for M1 support, would be good to have this PR sorted out.

kif:
I am pretty agnostic about this. A very long time ago (~10y), I had my projects in `project-src` folder and I was told that was not pythonic. Then we moved it to `project` with some mechanism preventing user from badly using it. Moving to `src/project` is neither better nor worse. It looks to me like a move back. Is it more pythonic today ? Go ahead if you wish.

t20100:
Apparently it is more pythonic today, so I'm in favor of going ahead.... and going back ahead in 10y.

vallsv:
Smaller the `setup.py`,  better we are.
If you think the change is safe, i think it's a good idea, already done with silx.

vallsv (approved):

t20100:
> If you think the change is safe, i think it's a good idea, already done with silx.

As long as tests pass, I don't see any issue here.

t20100:
Well, to confident...it broke debian packaging: see #118

PR #120, opened 2021-06-10
fix(LZ4): compression rate

Willian-Zhang:
this is a temporary fix for https://github.com/silx-kit/hdf5plugin/issues/121


t20100:
`LZ4_compress` is declared as obsolete in lz4: https://github.com/lz4/lz4/blob/355f60952938e5bd4d45118af720d4b8fb0c8a09/lib/lz4.c#L2454 so best not to use it.

t20100:
I'm closing this PR in favor of #123 which solves the issue without modifying the filter code.

Issue #122, opened 2021-06-10
ModuleNotFoundError: No module named 'hdf5plugin._config'


junyuewang:
My machine architechture is mips64el, system is Deepin based on Debian 10, I successfully built it by disabling sse2 and avx2 and changing flag -mcup to -march.
```
sudo python3 setup.py install
running install
creating /tmp/tmpb4wuc4uy/tmp
creating /tmp/tmpb4wuc4uy/tmp/tmpb4wuc4uy
mips64el-linux-gnuabi64-gcc -pthread -DNDEBUG -g -fwrapv -O2 -Wall -g -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -fPIC -c /tmp/tmpb4wuc4uy/source.cc -o /tmp/tmpb4wuc4uy/tmp/tmpb4wuc4uy/source.o -std=c++11
WARNING:__main__:SSE2 disabled: not available
WARNING:__main__:AVX2 disabled: not available
creating /tmp/tmp4p39ngq1/tmp
creating /tmp/tmp4p39ngq1/tmp/tmp4p39ngq1
mips64el-linux-gnuabi64-gcc -pthread -DNDEBUG -g -fwrapv -O2 -Wall -g -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -fPIC -c /tmp/tmp4p39ngq1/source.c -o /tmp/tmp4p39ngq1/tmp/tmp4p39ngq1/source.o -fopenmp
setup.py:88: DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead
  "CPU info detection does not support this architecture: SSE2 and AVX2 disabled")
WARNING:__main__:CPU info detection does not support this architecture: SSE2 and AVX2 disabled
INFO:__main__:Building with C++11: True
INFO:__main__:Building with native option: True
INFO:__main__:Building with SSE2: False
INFO:__main__:Building with AVX2: False
INFO:__main__:Building with OpenMP: True
running bdist_egg
running egg_info
writing src/hdf5plugin.egg-info/PKG-INFO
writing dependency_links to src/hdf5plugin.egg-info/dependency_links.txt
writing requirements to src/hdf5plugin.egg-info/requires.txt
writing top-level names to src/hdf5plugin.egg-info/top_level.txt
[...]

PR #126, opened 2021-06-11
Update setup.py to support mips64el

junyuewang:
For mips64el CPU, disable SSE2 and AVX2 in line 79,
and use -march=native in line 323.


kif:
LGTM

t20100:
related to #122

PR #127, opened 2021-06-12
Express Python version as an install requirement

mobiusklein:
Users who do not pin their version to a specific version installing from Python 2 will get an error when trying to install the library, rather than being directed to install an earlier version automatically by the package index.


kif:
LGTM

PR #129, opened 2021-06-14
Fixed typo in a test file name

t20100:
Fixed misspelling in test file: f**ic**decomp.h5 -> f**ci**decomp.h5


vasole:
I guess this should be part of #128

t20100:
That's a different subject, it makes review straight forward.

PR #116, opened 2021-05-27
Added ARM64 support; Added build config with environment variables

t20100:
This PR should add support for M1 by disabling SSE2 and AVX2 on `arm64` machines.
It also adds env var to configure default build options (useful with `cibuildwheel`)

~This is UNTESTED!~

PR #130, opened 2021-06-16
Changed build options override priority order

t20100:
This PR reworks the way build options (namely, `cpp11`, `sse2`, `avx2`, `native` and `openmp`)  are managed.

The current behaviour (since v3.0.0, PR #116) is:
- default values are `True` (except for `openmp` set to `False` on mac OS).
- `HDF5PLUGIN_*` environment variables overrides the defaults if they are set.
- Command line provided build options (e.g., `python setup.py build --sse2=False`) overrides defaults and env. vars
- After all that, options are checked against the current platform/compiler and disabled if not available.

This was making its best to avoid build failure by checking options at the end.
At the same time, this caused issues in supporting `arm64` and `mips64`, each time needing specific handling (see PR #116 and #126 ).

The proposed behaviour is:
- default values are probed from current os/platform/compiler, and if unknown all options but `cpp11` are disabled.
- `HDF5PLUGIN_*` env. vars overrides the defaults if they are set. 
- Command line provided build options (e.g., `python setup.py build --sse2=False`) overrides defaults and env. vars
- No checks (except for `native`, but that can be turned into an error).

So unknown platform should work straight away even if not at best performances.
Since there is no checks at the end, it is possible to force build options (except for `native`). At the same time it might fail if not supported.
It is now the user responsibility that provided build options are supported.

I also update `cpuinfo` to v8.0.0.

This needs quite a bit of testing on different platforms.

### DIFF
--- a/cpuinfo.py
+++ b/cpuinfo.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 # -*- coding: UTF-8 -*-
 
-# Copyright (c) 2014-2019, Matthew Brennan Jones <matthew.brennan.jones@gmail.com>
+# Copyright (c) 2014-2021 Matthew Brennan Jones <matthew.brennan.jones@gmail.com>
 # Py-cpuinfo gets CPU info with pure Python 2 & 3
 # It uses the MIT License
 # It is hosted at: https://github.com/workhorsy/py-cpuinfo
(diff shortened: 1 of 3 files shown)


t20100:
What do you think of the behaviour change?

Cautious review welcomed!

t20100:
Added a test to avoid the case: `sse2=False` and `avx2=True`.
In this case the blosc plugin (at least) is not working.

vasole:
I think OpenMP was giving some troubles having it enabled by default under windows.

t20100:
In current `main` branch, `OpenMP` is enabled by default on Windows (disabled on macos).
There is a compilation check that the compile and link `/openmp` flags are supported, so it will be disabled by default if it didn't managed to build with it.

But it is explicitly disabled when building wheels and conda packages.

t20100:
Reworked `setup.py` to separate:
- probing current host capabilities: `HostConfig` class
- build options: `BuildConfig` class with order of priority: 1. command line options, 2. env. var., 3. current host config.
- setuptools's `build` command

Also store filter shared lib file extension in configuration to know which file to load from `__init__.py` instead of using `glob`: There is a issue with `bdist_egg` (used by `setup.py install`) which creates some "stub" .py files next to the libs and then we tried to load those as shared lib...

t20100:
Ready for review, made probing of native more generic by using `cpuinfo` architecture regexp.
Tested on `aarch64` and partially on `mips`.

kif:
tested on obscure platforms...

PR #133, opened 2021-06-17
Improved build: Make sure to clean everything when changing the build config

t20100:


t20100:
closing as integrated in #130, since need for `pip install .` to work fine, else reuse code.

PR #136, opened 2021-06-23
Added detailled installation documentation

t20100:
This PR adds a documentation page on installation options.

PR #137, opened 2021-06-23
Added sphinx-based documentation

t20100:
This PR split the readme file into sphinx-based documentation.

This PR includes PR #136 about installation doc is merged.


t20100:
Added missing link to install doc in index.rst

t20100:
- Reworked to use `autodoc` to remove duplicate doc/docstring and updated content
- using readthedocs layout

Deployed here: http://www.silx.org/doc/hdf5plugin/latest/

I stop touching this!

andygotz:
@t20100 thanks for your excellent work packaging and documenting hdf5plugin! 

It deserves to be advertised some more - I will inform some of the communities I know.

t20100:
@andygotz I've added "supported architectures", thanks for the feedback.

vasole:
Perhpas we should add something in the lines of: 

Non h5py or non-python users can use the supplied libraries by setting the HDF5_PLUGIN_PATH environment variable to the output of the command `python -c  "import hdf5plugin; print(hdf5plugin.PLUGINS_PATH)"`

For instance, export HDF5_PLUGIN_PATH=$(python -c "import hdf5plugin; print(hdf5plugin.PLUGINS_PATH)") should allow MatLab or IDL users to read data compressed  using the supported plugins.

vasole:
Setting HDF5_PLUGIN_PATH would also allow already existing python code to read compressed data without any code modification.

t20100:
@vasole, indeed this was not documented, I added a paragraph about it building on your comments.

PR #147, opened 2021-07-07
Fixed debian packaging

### LINKED ISSUE #145: Tests are not run during debian/ubuntu packaging

### LINKED ISSUE #146: Look-up of plugin library is using too strict names
When packaged as debian/ubuntu packages, there is some extra characters between the lib name and le `.so` extension which leads to plugins not being registered (this does not occur with wheels).

t20100:
This PR:
- fixes `hdf5plugin` when packaged as debian/ubuntu package, closes #146
- updates debian 10/11 packaging by synchronizing config with what is used in silx, plus simplifying `rules` by removing overrides and using environment variables, and enable tests during packaging, closes #145


kif:
Looks god to me ... compiles on non-trivial architectures and tests run there

Issue #144, opened 2021-07-05
LZ4 from Intel's ipp ?


jonwright:
For the bitshuffle (LZ4) filter (on Intel) it seems the Intel IPP decompressor can be faster than the LZ4 which is currently bundled. 

The licensing situation is not entirely clear to me but I think Intel makes this code available as a binary blob in oneapi and also on [pypi ](https://pypi.org/project/ipp/). From https://software.intel.com/content/www/us/en/develop/articles/oneapi-commercial-faq.html : it says "Yes, all of the oneAPI Toolkits are available for free download and use for commercial and non-commercial purposes". Blosc also has an option to build against IPP, so presumably it is possible.

One option might be to build against a special LZ4 rather than the one which is bundled. There is a recipe from Intel for building a patched LZ4 library which calls their versions instead, from:  $(intel)/oneapi/ipp/latest/components/components/interfaces/ipp_lz4/readme.html
``` 
$ unzip lz4-1.9.3.zip
$ cd lz4-1.9.3
$ patch -p1 < ../lz4-1.9.3.patch.bin
```

Another method would be to send the calls to `LZ4_decompress_fast` and `LZ4_decompress_safe` to `DecodeLZ4` from ipp. This means locating the include+lib folders for ipp when compiling. Doing something like this in setup.py :
`extra_objects = [ 'path_to/libippdc.a', 'path_to/libippcore.a' ]` 
does not appear to introduce a new runtime dependency.

Before going much further into this I guess you want to see a benchmark to find out if it makes any difference and check it does not makes things worse for amd, etc ? Or maybe the binary blob already creates other problems ?

t20100:
Looks interesting.

As you say, it would need some benchmark since it would add some complexity to the compilation machinery and a dependency.

BTW, IPP is supported by c-blosc v2 but apparently not by c-blosc v1 which is the one used by the hdf5 filter, so it's not an option for blosc filter

t20100:
IPP licensing should not be an issue: we anyway won't bundle it nor redistribute binary wheels including IPP as long as it is a build from source option.

vasole:
Is this of renewed actuality now that Blosc 2 is added to the zoo?

jonwright:
[...]

PR #149, opened 2021-07-07
IPP benchmark

jonwright:
Here is a benchmark for issue #144.

You need  Intels IPP to use it (setup.py --ipp=...). At ESRF that can be `module load Intel/2020 ; echo $IPPROOT`, and then unload or get a fresh shell. Or download oneapi from Intel and install it somewhere. You only need IPP, but their compilers and openmp can also be useful. I think IPP also has patches for zfp, zlib, bzip2 as well ?

Gains vary, but look vaguely promising so far (e.g. two computers, seems to reproduce). I have the feeling it is not so good on real data but didn't track down why for the testcase yet. Could be diffraction peaks or something about the mask etc. It might also depend on exact versions of compressors.

Setting up a profile guided build using a benchmark might be interesting, but I would not want to try it via distutils.


t20100:
Thanks for the investigation!

We've so far avoided to patch the source code of the HDF5 filters, and I would prefer to keep it this way to avoid incompatibilities since there is a few copy of the code here and there (`hdf5plugin` being one more).

Yet, it is still possible to use IPP AND avoid modifying the filter code by compiling it with a `lz4.c` wrapper over IPP and doing the choice of the source file used at compile time.
As long as it is an opt-in (i.e., the user knows what he is doing), it would be fine this way.

Regarding performance gain, I just gave it a try on the `kevlar.h5` dataset available here: http://www.silx.org/pub/pyFAI/pyFAI_UM_2020/data_ID13 (1000 frames: 2167x2070 uint16, 1 chunk per frame) and couldn't see much performance gain, but that's just one benchmark.

Reading the whole `/entry/data/data` dataset at once:
* Without IPP:
  - OMP_NUM_THREADS=1: 7.74 s ± 59.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
  - OMP_NUM_THREADS=8: 4.55 s ± 81.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
* With IPP:
  - OMP_NUM_THREADS=1: 7.41 s ± 28 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
  - OMP_NUM_THREADS=8: 4.84 s ± 475 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

As for a profiling driven build process, it doesn't seems to worth the effort and it won't cover all tuning anyway: you'll still have to set `OMP_NUM_THREADS` to an appropriate value to get the most out of it.

jonwright:
OK - I will close this as it does not make sense to merge in the current state. Maybe the upstream code in setup.py can be changed to allow linking against a given lz4 library rather than the one that was vendored, probably better to see with them.

PR #150, opened 2021-07-20
Documentation: Added European HDF User Group presentation

t20100:
This PR adds the material presented at the European HDF User Group to the documentation of hdf5plugin.
It adds nbsphinx as dependency to build the doc from the notebook.


kif:
LGTM

Issue #153, opened 2021-10-13
[conda-forge] Test fails with hdf51.12.1 with python3.9 on power9


t20100:
See https://github.com/conda-forge/hdf5plugin-feedstock/pull/28 and https://github.com/conda-forge/hdf5plugin-feedstock/pull/29.

The issue can be reproduced locally by cloning https://github.com/conda-forge/hdf5plugin-feedstock and using the `build-locally.sh` script.

In this context, it seems the `pthread_atfork` symbol is missing when loading `libh5blosc.so`.

t20100:
should be closed by #154

PR #154, opened 2021-10-13
Fixed compilation issue occuring on ppc64le in conda-forge

t20100:
This PR aims at fixing the test failure on conda-forge python3.9/ppc64le (see #153).
`blosc` is using `pthread`, but this was not explicitly set as compile and link argument.

It also adds a `.gitmodules` file for snappy submodules since conda-forge runs `git submodule update --init --recurse` as part of its build process.

Similar issue of scipy (https://github.com/scipy/scipy/issues/11323) with similar solution (https://github.com/scipy/scipy/pull/11324).


kif:
LGTM

PR #156, opened 2021-10-14
Downgraded snappy to 1.1.8

t20100:
There is an issue building snappy 1.1.9 on macos 10.12 so downgrading to 1.1.8
See https://github.com/google/snappy/pull/142


kif:
LGTM

PR #167, opened 2022-05-17
Added pyproject.toml

### LINKED ISSUE #131: Add pyproject.toml
The currently recommended python packaging way is to provide a `pyproject.toml` (PEP 517 and PEP518)... and I guess not having it will stop to work at some point.

But when adding `pyproject.toml`, build options does not work any longer, only env. var. are working.

t20100:
This PR adds the missing `pyproject.toml` file to the project.
It also deprecates build options passed through the command line since calling `setup.py` directly is now deprecated.

closes #131

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -302,25 +302,12 @@ class Build(build):
     """Build command with extra options used by PluginBuildExt"""
 
     user_options = [
-        ('hdf5=', None, "Custom path to HDF5 (as in h5py). "
-         "Default: HDF5PLUGIN_HDF5_DIR environment variable if set."),
-        ('openmp=', None, "Whether or not to compile with OpenMP. "
-         "Default: HDF5PLUGIN_OPENMP env. var. if set, else "
-         "True if probed (always False on macOS)."),
-        ('native=', None, "True to compile specifically for the host, "
-         "False for generic support (For unix compilers only). "
-         "Default: HDF5PLUGIN_NATIVE env. var. if set, else "
-         "True on supported architectures, False otherwise"),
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -52,6 +52,13 @@
     logger.warning("Architecture is not supported by cpuinfo")
     cpuinfo = None
 
+try:  # Embedded copy of cpuinfo
+    from cpuinfo import _parse_arch as cpuinfo_parse_arch
+except Exception:
+    try:  # Installed version of cpuinfo (when installing with pip)
+        from cpuinfo.cpuinfo import _parse_arch as cpuinfo_parse_arch
```
t20100, line 59: For whatever reason when running `pip install .`, the `cpuinfo.py` module at the root is not found.
Since it is installed through the `pyproject.toml` file, let's use it.

t20100:
I deprecated the command line `setup.py`. They can still be used, but I'd expect all the update of `setuptools` and co to break it sooner or later.

vasole:
In general I do not like to see things like ``py-cpuinfo==8.0.0`` Is it possible to write ``py-cpuinfo>=8.0.0``?

t20100:
I prefer to use `py-cpuinfo==8.0.0` here since it uses a private function of this package... and I don't want to watch every releases of `py-cpuinfo`.

It is only needed for building from source with `pip install` and it gets installed in a virtualenv by pip that is used only for the build.
Other install methods don't install/use it: wheels and conda packages do not rely on this and `python setup.py` will use the local copy of it.
So this version pin-pointing should not have any impact on the user installation.

vasole (approved):

PR #169, opened 2022-05-17
Added `config.embedded_filters` to advertise embedded filters, and HDF5PLUGIN_STRIP to configure it

### LINKED ISSUE #162: Make hdf5plugin friendly for Linux packaging
`hdf5plugin` was initially made to provide HDF5 plugins when the ones from the system cannot be used (in virtual env., or not available).
However, having the possibility to package `hdf5plugin` for Linux distribution would allow to also use the `hdf5plugin` API to configure the filters.

This means allowing to strip all/some filters when building the package and instead relying on those from the system and maybe to adapt the initialisation.

t20100:
This PR adds a build time option to remove some/all hdf5 compression filters from the build of hdf5plugin.
This provides a way for packagers to bundle the compression filters separately, yet providing hdf5plugin API.
This is controlled through the undocumented `HDF5PLUGIN_STRIP` env. var.

It also provides the list of embedded filters through `hdf5plugin.config.embedded_filters`. It is generated from HDF5PLUGIN_STRIP and also when building without C++11 where fcidecomp cannot be compiled.

closes #162

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -261,6 +261,15 @@ class BuildConfig:
             compile_args.extend(host_config.native_compile_args)
         self.__compile_args = tuple(compile_args)
 
+        self.embedded_filters = []
+
+        self.not_embedded_filters = []
+        env_strip = os.environ.get('HDF5PLUGIN_STRIP', None)
+        if env_strip:
+            self.not_embedded_filters = [
+                name.strip().lower() for name in env_strip.split(',')
+            ]
+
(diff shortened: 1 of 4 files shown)


vasole (approved):

t20100:
I just improved the way to check for test to skip: If the filter is "embedded" the test is run even if the filter is not available from libhdf5...
This is the case if there is an issue loading the filter and we really want the tests to cover that!

PR #171, opened 2022-05-18
Updated Bitshuffle filter to v0.4.2, Bitshuffle API change with deprecation warning

### LINKED ISSUE #170: Update bitshuffle
There is a new release of bitshuffle: https://github.com/kiyo-masui/bitshuffle/releases/tag/0.4.2
It adds zstd compression as an alternative to lz4.

t20100:
This PR updates bitshuffle to the latest version 0.4.2 which adds zstd as an alternative to lz4.
It is built with the version of zstd included in blosc.

The `Bitshuffle` helper class API was not suited for supporting an extra compression, so I changed it to match that of the `Blosc` helper. This deprecates the previous API with `lz4=True|False` argument, but it should remain fully compatible.

closes #170


t20100:
There is a build issue on Windows, PR https://github.com/kiyo-masui/bitshuffle/pull/122 fixes it.

vasole:
There is a compilation error under windows. Should I try to take care of it?

t20100:
There is a PR in bitshuffle for it: https://github.com/kiyo-masui/bitshuffle/pull/122 but it's pending review.

vasole:
I think we should fix it in our code in the mean time. If people use it under linux, they will generate files not usable under windows...

t20100:
I'd prefer to wait for the fix to be accepted upstream first.

But yes, until then, HDF5 files created with the HDF5 filter from `bitshuffle` Python package and using `zstd` are not supported by `hdf5plugin`.

t20100:
I added the patch from https://github.com/kiyo-masui/bitshuffle/pull/122 and documented it.
I'm no fan of patching the filters, but here it's only for solving a build issue.

vasole (approved):

Issue #110, opened 2021-03-01
Adding BZIP2 and MAFISC filters


t20100:
Once JPEG is added (#85), adding  [BZIP2](https://portal.hdfgroup.org/display/support/Filters#Filters-307) and [MAFISC](https://portal.hdfgroup.org/display/support/Filters#Filters-32002) would make `hdf5plugin` embedding all "commonly used Registered Filter Plugins [..] packaged into the HDF5 Plugin software" (see [HDF5 Filter Plugins](https://portal.hdfgroup.org/display/support/HDF5+Filter+Plugins): BZIP2, JPEG, LZF, BLOSC, MAFISC, LZ4, Bitshuffle, and ZFP) except LZF which is already provided by `h5py`.

vasole:
BZip2 already in.

t20100:
There was initially no request for adding this filter and it is currently not in https://github.com/HDFGroup/hdf5_plugins, so I propose to close this issue.

PR #182, opened 2022-06-03
Added BZip2 compression filter

t20100:
This PR adds the BZip2 compression filter from `PyTables` to `hdf5plugin`.
It also updates slightly the contribution documentation.

I didn't use `git subtree` for PyTables, instead I copy-pasted the 3 relevant files while keeping the same folder structure. As we add filters, we're also adding a lot of useless files, so maybe time to change the strategy of embedding everything, what do you think of this?

I'm also missing a hdf5 file with bzip2 compression that was produced by another tool.

Related to #110


t20100:
Let's keep this as a Draft PR for now, and wait for an effective need to add it to `hdf5plugin`.

vasole:
For easy-to-handle filters I do not mind to add them to the code repository. If PyTables uses it, I would add it.

t20100:
OK, up to you then!
My feeling was that we'll have to maintain a filter that we are not sure is used, but as you say it's a simple one.

t20100:
OK, up to you then!
My feeling was that we'll have to maintain a filter that we are not sure is used, but as you say it's a simple one.

PR #187, opened 2022-10-18
Added `get_config()` function to retrieve information

### LINKED ISSUE #185: Improve logging/information about `hdf5plugin` loaded filters
Proposition from @aragilar (https://github.com/h5py/h5py/issues/2163#issuecomment-1280714805):

> hdf5plugin does appear to log almost all of what it does with respect to loading filters (the only bit missing would be an explicit "hdf5plugin has loaded <plugin> at <path>")

> On the hdf5plugin side, I could see adding a debug_info() function (feel free to choose a different name) which gives details of what it has done (e.g. what version it is, what are versions of the filters its loaded, what failed) as a possible improvement?

t20100:
This PR:
- adds a debug log when successfully registering a filter.
- adds and document a `get_config()` function with provides access to both build-time options and runtime registered filters.

Previously `hdf5plugin.config` was available (and still is) to provide build time config but it is not documented.

I hesitated between providing a dedicated `get_registered_filters()` function and providing a more general `get_config()` one as this PR proposes. I chose the later to avoid having too many functions for accessing debug/more advanced information. Any other opinion? Better naming?

closes #185

### DIFF
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -6,11 +6,6 @@
 
 ``hdf5plugin`` allows using additional HDF5 compression filters with `h5py`_ for reading and writing compressed datasets.
 
-Available constants:
-
-* ``hdf5plugin.FILTERS``: A dictionary mapping provided filters to their ID
-* ``hdf5plugin.PLUGIN_PATH``: The directory where the provided filters library are stored.
-
 Read compressed datasets
 ++++++++++++++++++++++++
 
@@ -101,6 +96,21 @@ Zstd
 .. autoclass:: Zstd
    :members:
    :undoc-members:
+
+Get information about hdf5plugin
+++++++++++++++++++++++++++++++++
+
+Available constants:
+
+.. autodata:: FILTERS
+   :annotation:
+
+.. autodata:: PLUGIN_PATH
+   :annotation:
+
+Functions:
+
+.. autofunction:: get_config
 
 Use HDF5 filters in other applications
 ++++++++++++++++++++++++++++++++++++++
--- a/src/hdf5plugin/__init__.py
+++ b/src/hdf5plugin/__init__.py
@@ -450,6 +450,7 @@ def _init_filters():
         # Skip filters that were not embedded
         if name not in config.embedded_filters:
             _logger.debug("%s filter not available in this build of hdf5plugin.", name)
+            yield name, ("unknown", "unknown")
             continue
 
         # Check if filter is already loaded (not on buggy HDF5 versions)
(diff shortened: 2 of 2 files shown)


vasole:
Why is the information about the constants missing? I refer to:

* ``hdf5plugin.FILTERS``: A dictionary mapping provided filters to their ID
* ``hdf5plugin.PLUGIN_PATH``: The directory where the provided filters library are stored.

The explanation of the constants seems to have disappeared with the modifications. I think it is particularly needed in order to make clear that PLUGIN_PATH is the directory containing the filters supplied by hdf5plugin and not the HDF5_PLUGIN_PATH

vasole:
The situation is:

- we need to know how `hdf5plugin` was configured
- we need to know what filters have been registered
- we need to know what registered filters come form `hdf5plugin` and which ones do not

t20100:
Yes, the constants are now documented with `autodoc` to avoid duplication between the docstring and the rst.

vasole:
I agree it is a more generic functionality to retrieve all filters.

However, in the case we have skipped a filter because it was already loaded, I think we should somehow let the user retrieve that information without having to toggle the logging level and (perhaps) restart again.

@aragilar suggested to provide hdf5plugin loaded filter and used path (obviously if we load it we use `hdf5plugin.PLUGIN_PATH`) but if we do not load it because already there then the path is `unknown`...

vasole:
Basically I am suggested to add `yield name, ("unknown", "unknown")` after  `_logger.info("%s filter already loaded, skip it.", name)`

t20100:
Yes I agree.

As it is, one can already find this information by checking both `embedded_filters` and `registered_filters`, but all this information can be included in `registered_filters`.

t20100:
I added a status message to the `registered_filters`:
To provide something like this:
```python
In [1]: import hdf5plugin

In [2]: hdf5plugin.get_config().registered_filters
Out[2]: 
{'blosc': FilterStatus(status='Missing: not embedded in hdf5plugin', filename=None),
 'bshuf': FilterStatus(status='Missing: not embedded in hdf5plugin', filename=None),
 'bzip2': FilterStatus(status='Missing: not embedded in hdf5plugin', filename=None),
 'lz4': FilterStatus(status='Registered', filename='/hdf5plugin/plugins/libh5lz4.dylib'),
 'zfp': FilterStatus(status='Missing: not embedded in hdf5plugin', filename=None),
 'zstd': FilterStatus(status='Missing: not embedded in hdf5plugin', filename=None),
 'fcidecomp': FilterStatus(status='Missing: not embedded in hdf5plugin', filename=None)}
```

but it's maybe too much, this is already in the logging message, maybe better to keep it simple.

vasole:
I think the requests are already fulfilled. The rest is just a question of taste.

PR #188, opened 2022-10-19
Build: Updated `HDF5PLUGIN_STRIP` behaviour and added support for `"all"` value

t20100:
This PR is a follow-up of PR #169 for stripping compression filters from the package at build time.

It adds:
- filtering out of C libraries from the build when they correspond to stripped filters.
- support for a `"all"` value which strips all filters at once.

Filtering out stripped plugins is moved from inside the `build` command to before calling `setup()` since there is no need to be inside the command to do it.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -270,13 +270,6 @@ class BuildConfig:
 
         self.embedded_filters = []
 
-        self.not_embedded_filters = []
-        env_strip = os.environ.get('HDF5PLUGIN_STRIP', None)
-        if env_strip:
-            self.not_embedded_filters = [
-                name.strip().lower() for name in env_strip.split(',')
-            ]
-
     hdf5_dir = property(lambda self: self.__hdf5_dir)
     filter_file_extension = property(lambda self: self.__filter_file_extension)
     use_cpp11 = property(lambda self: self.__use_cpp11)
@@ -365,18 +358,10 @@ class Build(build):
                 (name, info) for name, info in self.distribution.libraries
                 if '-std=c++11' not in info.get('cflags', [])]
 
-            # Add C++11-only extensions to stripped plugins
-            for ext in self.distribution.ext_modules:
-                if isinstance(ext, HDF5PluginExtension) and ext.cpp11_required:
-                    self.hdf5plugin_config.not_embedded_filters.append(
-                        ext.hdf5_plugin_name
-                    )
-
-        # Filter out stripped plugins
-        self.distribution.ext_modules = [
-            ext for ext in self.distribution.ext_modules
-            if not isinstance(ext, HDF5PluginExtension) or ext.hdf5_plugin_name not in self.hdf5plugin_config.not_embedded_filters
-        ]
+            # Filter out C++11-only extensions
+            self.distribution.ext_modules = [
+                ext for ext in self.distribution.ext_modules
+                if not (isinstance(ext, HDF5PluginExtension) and ext.cpp11_required)]
 
         self.hdf5plugin_config.embedded_filters = [
             ext.hdf5_plugin_name for ext in self.distribution.ext_modules
@@ -548,6 +533,10 @@ def prefix(directory, files):
     return ['/'.join((directory, f)) for f in files]
 
 
+PLUGIN_LIB_DEPENDENCIES = dict()
+"""Mapping plugin name to library name they depend on"""
+
+
 # bitshuffle (+lz4) plugin
 # Plugins from https://github.com/kiyo-masui/bitshuffle
 bithsuffle_dir = 'src/bitshuffle'
@@ -666,6 +655,8 @@ blosc_plugin = HDF5PluginExtension(
     avx2=avx2_kwargs,
     cpp11=cpp11_kwargs,
     )
+PLUGIN_LIB_DEPENDENCIES['blosc'] = 'snappy'
+
 
 # HDF5Plugin-Zstandard
 zstandard_dir = os.path.join("src", "HDF5Plugin-Zstandard")
@@ -757,6 +748,8 @@ fcidecomp_plugin = HDF5PluginExtension(
     cpp11_required=True,
     define_macros=[('CHARLS_STATIC', 1)],
     )
+PLUGIN_LIB_DEPENDENCIES['fcidecomp'] = 'charls'
+
 
 # CharLS
 charls_dir = "src/charls/src"
@@ -789,6 +782,7 @@ h5zfp_plugin = HDF5PluginExtension(
     extra_compile_args=extra_compile_args,
     extra_link_args=extra_link_args,
     )
+PLUGIN_LIB_DEPENDENCIES['zfp'] = 'zfp'
 
 # zfp
 zfp_dir = os.path.join("src", "zfp")
@@ -800,16 +794,47 @@ zfp_lib = ('zfp', {
     'cflags': ['-DBIT_STREAM_WORD_TYPE=uint8'],
     })
 
-libraries = [snappy_lib, charls_lib, zfp_lib]
 
-extensions = [bzip2_plugin,
-              lz4_plugin,
-              bithsuffle_plugin,
-              blosc_plugin,
-              fcidecomp_plugin,
-              h5zfp_plugin,
-              zstandard_plugin,
-              ]
+def apply_filter_strip(libraries, extensions, dependencies):
+    """Strip C libraries and extensions according to HDF5PLUGIN_STRIP env. var."""
+    stripped_filters = set(
+        name.strip().lower()
+        for name in os.environ.get('HDF5PLUGIN_STRIP', '').split(',')
+        if name.strip()
+    )
+
+    if 'all' in stripped_filters:
+        return [], []
+
+    # Filter out library that won't be used because of stripped filters
+    lib_names = set(
+        lib_name for filter_name, lib_name in dependencies.items()
+        if filter_name not in stripped_filters
+    )
+    libraries = [
+        lib for lib in libraries if lib[0] in lib_names
+    ]
+
+    # Filter out stripped filters
+    extensions = [
+        ext for ext in extensions
+        if isinstance(ext, HDF5PluginExtension) and  ext.hdf5_plugin_name not in stripped_filters
+    ]
+    return libraries, extensions
+
+libraries, extensions = apply_filter_strip(
+    libraries=[snappy_lib, charls_lib, zfp_lib],
+    extensions=[
+        bzip2_plugin,
+        lz4_plugin,
+        bithsuffle_plugin,
+        blosc_plugin,
+        fcidecomp_plugin,
+        h5zfp_plugin,
+        zstandard_plugin,
+    ],
+    dependencies=PLUGIN_LIB_DEPENDENCIES,
+)
 
 
 # setup


vasole:
So, the goal is to reduce hdf5plugin to be just an API.

The Python developers can safely use `hdf5plugin` it in their programs without having to care if the libraries have been provided by their linux distribution or us.

vasole (approved):
LGTM

bnavigator:
This is nice, but how do I actually provide the filter plugins for system provided libraroes then? Currently it's either embedded with a static library or no plugin at all.

t20100:
> how do I actually provide the filter plugins for system provided libraries then? Currently it's either embedded with a static library or no plugin at all.

This project primary goal is to provide hdf5 filters for h5py in a Python environment, not to provide system-wide hdf5 filters.

However, provided you have installed `hdf5plugin`, you can run the following to make the filters available for other applications **for reading only**:

```python
export HDF5_PLUGIN_PATH=`python3 -c "
import hdf5plugin; print(hdf5plugin.PLUGINS_PATH)"`
```

If you want to install hdf5 filters system-wide, you need to check what is available for your OS and otherwise need to compile filters for the libhdf5 installed by your OS.

BTW embedded filter libraries are dynamic libraries.

bnavigator:
> However, provided you have installed hdf5plugin, you can run the following to make the filters available for other applications for reading only

That is a different use case. I want the API calls with external filters not the embedded filters for external programs.

Specifically, I want python-fabio to use the bitshuffle filter, which it expects to get from hdf5plugin, but could also be provided directly from the bitshuffle package.

> BTW embedded filter libraries are dynamic libraries.

Yet all the dynamic libraries are compiled from the bundled sources and there is no config option to directly link to existing shared libs, is there?

t20100:
hdf5plugin do not register its embedded filters if there is one already available:

https://github.com/silx-kit/hdf5plugin/blob/2782806b8e1eb82f7b78dc2075296cf132a74a40/src/hdf5plugin/_utils.py#L237-L260

So, if you want to use the one embedded in `bitshuffle`, you have to register it to libhdf5 before importing `hdf5plugin`.

If you want to make sure bitshuffle is not embedded, you can build hdf5plugin from source with the environmnent variable: `HDF5PLUGIN_STRIP=bshuf` or `HDF5PLUGIN_STRIP=all`.

bnavigator:
That's not what I want. I want to package hdf5plugin to have and register the filters so they can be used by packages like fabio without having to modify the latter.

But I do not want hdf5plugin to create big "dynamic" .so libraries which statically link vendored compression libraries. I  want it to have slim filter plugins dynamically linking to the shared libraries already available on the distribution.

Compare your published wheels with https://build.opensuse.org/package/show/devel:languages:python:numeric/python-hdf5plugin.

Your wheel has an SZ3 plugin that is 100MB thick.

```
[ben@skylab:~/tmp/hdf5wheelenv]% cat lib64/python3.11/site-packages/hdf5plugin-4.4.0.dist-info/WHEEL                                                                                                     [0]
Wheel-Version: 1.0
Generator: bdist_wheel (0.42.0)
Root-Is-Purelib: false
Tag: py3-none-manylinux_2_17_x86_64
Tag: py3-none-manylinux2014_x86_64
[ben@skylab:~/tmp/hdf5wheelenv]% ls lib64/python3.11/site-packages/hdf5plugin/plugins -lh                                                                                                                [0]
insgesamt 175M
-rwxr-xr-x 1 ben users  17M  4. Mär 18:05 libh5blosc2.so
-rwxr-xr-x 1 ben users  14M  4. Mär 18:05 libh5blosc.so
-rwxr-xr-x 1 ben users  11M  4. Mär 18:05 libh5bshuf.so
-rwxr-xr-x 1 ben users 332K  4. Mär 18:05 libh5bzip2.so
-rwxr-xr-x 1 ben users 2,9M  4. Mär 18:05 libh5fcidecomp.so
-rwxr-xr-x 1 ben users 1,2M  4. Mär 18:05 libh5lz4.so
-rwxr-xr-x 1 ben users 102M  4. Mär 18:05 libh5sz3.so
-rwxr-xr-x 1 ben users  16M  4. Mär 18:05 libh5sz.so
-rwxr-xr-x 1 ben users 2,3M  4. Mär 18:05 libh5zfp.so
-rwxr-xr-x 1 ben users 9,7M  4. Mär 18:05 libh5zstd.so
[ben@skylab:~/tmp/hdf5wheelenv]% ldd lib64/python3.11/site-packages/hdf5plugin/plugins/libh5sz3.so                                                                                                       [0]
        linux-vdso.so.1 (0x00007ffd55cd4000)
        libstdc++.so.6 => /lib64/libstdc++.so.6 (0x00007fec0c200000)
        libm.so.6 => /lib64/libm.so.6 (0x00007fec0c519000)
        libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007fec0d0aa000)
        libpthread.so.0 => /lib64/libpthread.so.0 (0x00007fec0d0a3000)
        libc.so.6 => /lib64/libc.so.6 (0x00007fec0be00000)
        /lib64/ld-linux-x86-64.so.2 (0x00007fec0d100000)
[ben@skylab:~/tmp/hdf5wheelenv]% ldd lib64/python3.11/site-packages/hdf5plugin/plugins/libh5blosc2.so                                                                                                    [0]
        linux-vdso.so.1 (0x00007fff1dbc6000)
        libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f593bb91000)
        libc.so.6 => /lib64/libc.so.6 (0x00007f593b800000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f593bd78000)
```

When I ruthlessly patch out the compilation and static linking of the libraries and use system libraries instead, I get a much slimmer package, which can leverage the system libraries, where I have them available:

```
[ben@skylab:…umeric/python-hdf5plugin]% ls -lh /usr/lib64/python3.11/site-packages/hdf5plugin/plugins/                                                                                                   [0]
insgesamt 9,9M
-rwxr-xr-x 1 root root  31K  5. Mär 10:04 libh5blosc2.so
-rwxr-xr-x 1 root root  23K  5. Mär 10:04 libh5blosc.so
-rwxr-xr-x 1 root root  47K  5. Mär 10:04 libh5bshuf.so
-rwxr-xr-x 1 root root  23K  5. Mär 10:04 libh5bzip2.so
-rwxr-xr-x 1 root root 231K  5. Mär 10:04 libh5fcidecomp.so
-rwxr-xr-x 1 root root 7,9M  5. Mär 10:04 libh5sz3.so
-rwxr-xr-x 1 root root 1,4M  5. Mär 10:04 libh5sz.so
-rwxr-xr-x 1 root root 347K  5. Mär 10:04 libh5zfp.so
[ben@skylab:…umeric/python-hdf5plugin]% ldd /usr/lib64/python3.11/site-packages/hdf5plugin/plugins/libh5sz3.so                                                                                           [0]
        linux-vdso.so.1 (0x00007ffc18197000)
        libzstd.so.1 => /lib64/glibc-hwcaps/x86-64-v3/libzstd.so.1.5.5 (0x00007fa415c31000)
        libstdc++.so.6 => /lib64/libstdc++.so.6 (0x00007fa415000000)
        libm.so.6 => /lib64/libm.so.6 (0x00007fa415319000)
        libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007fa415c02000)
        libc.so.6 => /lib64/libc.so.6 (0x00007fa414c00000)
        /lib64/ld-linux-x86-64.so.2 (0x00007fa415d07000)
[ben@skylab:…umeric/python-hdf5plugin]% ldd /usr/lib64/python3.11/site-packages/hdf5plugin/plugins/libh5blosc2.so                                                                                        [0]
        linux-vdso.so.1 (0x00007ffc540ff000)
        libblosc2.so.2 => /lib64/glibc-hwcaps/x86-64-v3/libblosc2.so.2.13.1 (0x00007f6863f0d000)
        libc.so.6 => /lib64/libc.so.6 (0x00007f6863c00000)
        libz.so.1 => /lib64/glibc-hwcaps/x86-64-v3/libz.so.1.3 (0x00007f6863ef3000)
        libzstd.so.1 => /lib64/glibc-hwcaps/x86-64-v3/libzstd.so.1.5.5 (0x00007f6863e44000)
        liblz4.so.1 => /lib64/glibc-hwcaps/x86-64-v3/liblz4.so.1.9.4 (0x00007f6863e20000)
        libm.so.6 => /lib64/libm.so.6 (0x00007f6863b19000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f6863fb7000)
```

t20100:
I'm not sure to understand what you want. Do you want to package hdf5plugin for Linux?

The purpose of the embedded filters is the same as Python wheels: to be as standalone as possible... and so "big". Definitely not what you want for Linux packaging.

You probably want to set  `HDF5PLUGIN_STRIP=all` and to package the hdf5 filters separately like bitshuffle (so they can be used from non-Python software too). This is what is done for Debian.

As it is, there is no way to custom what is/is not included in the embedded filters.

bnavigator:
The problem with that approach is, that I would have to modify packages which have hdf5plugin in their dependencies. (Debian did not update fabio yet: https://salsa.debian.org/science-team/python-fabio)

I would not have asked if it wasn't for this comment from @vasole  (https://github.com/silx-kit/hdf5plugin/pull/188#issuecomment-1285313681):
> The Python developers can safely use hdf5plugin it in their programs without having to care if the libraries and plugins have been provided by their linux distribution or us.

If you have to register them separately, that defies the purpose of not "having to care".

vasole:
> If you have to register them separately, that defies the purpose of not "having to care".

If your distribution provides the plugins, hdf5plugin will detect they are loaded and it will not use those supplied.

So, the end user does not care if the plugins are provided by hdf5plugin or the system. What you cannot do is to strip them from hdf5plugin because then *you* have to make sure they are available from the system.

bnavigator:
> unless *you* make sure they are available from the system.

> The Python developers can safely use hdf5plugin it in their programs without having to care

In my opinion, these statements contradict each other. There is no way to convince a python package to register a filter next to hdf5plugin without transitioning from package maintainer into the developer role. Because you have to modify the python code:

> So, if you want to use the one embedded in bitshuffle, you have to register it to libhdf5 before importing hdf5plugin.

t20100:
> I would have to modify packages which have hdf5plugin in their dependencies.

Why?
If hdf5 filters are packaged separately, you can add them as dependencies of the hdf5plugin package and then fabio would not have to care.

> What you cannot do is to strip them from hdf5plugin unless you make sure they are available from the system.
> Debian packagers strip them from hdf5plugin but they make the plugins available via the system.

To be correct, you can, the filters will just not be available.
In Debian, no filter is embedded in hdf5plugin and not all filters are available in seperate packages.

bnavigator:
> If hdf5 filters are packaged separately, you can add them as dependencies of the hdf5plugin package and then fabio would not have to care.

I did that, it did not work, because hdf5plugin still tells fabio that there is no filter.
- Build hd5plugin with HD5PLUGIN_STRIP=all
- Build bitshuffle
- Run fabio test suite (or try to read a bitshuffle h5 manually)


Let me try to reproduce it...

t20100:
There is different ways to register hdf5 filters:
- Install the filter in the folder libhdf5 is looking for them. In this case they are automatically loaded. This is where I would expect Linux packages to install the hdf5 filters so they are available system-wide.
- Register the filter with the libhdf5 API: This is what hdf5plugin is doing.

bitshuffle actually provides both:
- The Debian package provides the filters in the libhdf5 folder (See https://packages.debian.org/bookworm/amd64/bitshuffle/filelist)
- `import bitshuffle.h5` registers the filter through the API (See https://github.com/kiyo-masui/bitshuffle/blob/b9a1546133959298c56eee686932dbb18ff80f7a/bitshuffle/h5.pyx#L93C5-L99)

t20100:
> Linux packaging, see the opensuse link I posted above.

Thanks for the packaging effort!

A few relevant links:
- [Official list of registered HDF5 filters](https://portal.hdfgroup.org/documentation/hdf5-docs/registered_filter_plugins.html). Unfortunately it no longer provide links to the filters source code...
- [List of upstream repository of filters embedded in hdf5plugin](https://hdf5plugin.readthedocs.io/en/stable/information.html#hdf5-filters-and-compression-libraries)
- [Fork of hdf5 filters by the hdfgroup](https://github.com/HDFGroup/hdf5_plugins), not all filters are up-to-date (e.g., bitshuffle).

To build system-wide hdf5 filters, you should not build them through `hdf5plugin` because we add [a C file](https://github.com/silx-kit/hdf5plugin/blob/main/src/hdf5_dl.c) to [workaround linking issues](https://hdf5plugin.readthedocs.io/en/stable/hdf5plugin_EuropeanHUG2021.html#The-Problem) when using h5py wheel.

Having a look at `hdf5plugin-system-libs.patch`, you can spare patching the SIMD support by using environment variables: https://hdf5plugin.readthedocs.io/en/stable/install.html#available-options

Issue #186, opened 2022-10-18
Build issue on macos12.6


t20100:
Failed to build with macos12.6 on a x86_64 CPU, python3.10 with default build options.

First error:
```
clang: error: the clang compiler does not support '-march=native'
```
This option was apparently not available for clang on ARM before v15.0.0

Using `HDF5PLUGIN_NATIVE=False` fixes it error, but build fails later with:
```
src/c-blosc/blosc/bitshuffle-avx2.c:22:4: error: AVX2 is not supported by the target architecture/platform and/or this compiler.
...
```
This seems to be due to building fat library (`-arch arm64 -arch x86_64` options) with initial option detection being run only on host architecture.

PR #189, opened 2022-10-19
Build: Added check of native flags availability

t20100:
This PR adds an effective check of native compilation flags availability because it is not available for universal2 build on macos.

Related to #186


vasole (approved):
LGTM

vasole:
What has happened to the CI in this request?

PR #191, opened 2022-10-21
Updated logging, code clean-up

t20100:
This PR:
- Provides a small update for the logging by providing the traceback in case of catched exception
- Removes unsued `registered_filters` variable
- Replaces `"unknown"` string by `None` for the ctypes lib object


vasole (approved):
LGTM

PR #192, opened 2022-10-21
Updated `snappy` to v1.1.9 (used by the `blosc` filter)

t20100:
This PR updates the version of the `snappy` compression library used by `blosc` to the latest available version and update the documentation accordingly.


vasole (approved):
OK, just documentation from our side.

PR #200, opened 2022-11-08
Updated `c-blosc` to commit 9dc93b1 and `zstd` to v1.5.2

### LINKED ISSUE #186: Build issue on macos12.6

t20100:
This PR updates `c-blosc` to commit 9dc93b1 initially to fix macos `universal2` build.
It also updates `zstd` to v1.5.2 which includes a piece of x86_64 assembly code (that looks to work by passing it as `extra_objects` to the `Extension`s).
This still needs testing on architecture different from x86_64.

closes #186

### DIFF
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -53,6 +53,9 @@ Available options
    * - ``HDF5PLUGIN_AVX2``
      - Whether or not to compile with `AVX2`_ support. avx2=True requires sse2=True.
        Default: True on x86 when probed, False otherwise
+   * - ``HDF5PLUGIN_BMI2``
+     - Whether or not to compile Zstandard with `BMI2`_ support if available.
+       Default: True on Linux and MacOS, False otherwise
    * - ``HDF5PLUGIN_CPP11``
      - Whether or not to compile C++11 code if available.
        Default: True if probed.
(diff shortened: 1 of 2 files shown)


t20100:
The assembly code is only enabled under linux and macos and explicitly disabled otherwise.
There is further checks with C macros to check if it is actually there.

Tested also on Power9 and macos ARM.

One issue may be for generic builds (manylinux, conda): This assembly code uses [BMI2](https://en.wikipedia.org/wiki/X86_Bit_manipulation_instruction_set#BMI2_(Bit_Manipulation_Instruction_Set_2)) instructions which are apparently available starting from [Haswell](https://en.wikipedia.org/wiki/Haswell_(microarchitecture)) Intel processors and [Excavator](https://en.wikipedia.org/wiki/Excavator_(microarchitecture)) AMD processors, so available since about 10 years.

t20100:
I added an env. var. `HDF5PLUGIN_BMI2=True|False` (default `True`) to allow disabling this.

t20100:
> I like the conservative approach concerning assembler code and windows

If you know how to handle assembly code in build through `setuptools` on Windows, I'd take it.

t20100:
From a quick look (I haven't tested anything), Visual Studio does not seems to support assembly code by default.
From what I've seen, one need to install an extra tool like [nasm](https://www.nasm.us/) to compile the assembly code first.
With `gcc` and `clang`, one can pass the `.S` file directly at link time: thus the use of `extra_objects` in this PR and that I disabled it elsewhere including Windows.
We can consider enabling it for Windows when it is not using Visual Studio, but it is really a corner case and not worth the effort to me.

PR #203, opened 2022-11-10
Adding SZ HDF5 filter

orioltinto:
Right now the filter works in Ubuntu but haven't tried anywhere else.

The implementation seems clean enough to me except for the workaround that I had to do for running the **configure script**:

```python
            if lib_name == "sz":
                from pathlib import Path
                # Execute configure script
                configure_path = Path().cwd() / "src" / "SZ" / "configure"
                self.spawn([configure_path.as_posix()])
                # Move config.h from the current working directory to SZ directory.
                config_h_path = Path().cwd() / "config.h"
                config_h_path.rename(configure_path.parent / "config.h")
```
If anybody finds a better solution it would be great, otherwise this works.


vasole:
To give a chance of having something running under windows, I would certainly prefer to use cmake rather than configure.

At least it is what I have used to generate config.h

vasole:
Your changes approved. Work continues on #206

PR #205, opened 2022-11-22
Split `__init__.py` into submodules

### LINKED ISSUE #202: Restructure `hdf5plugin/__init__.py`
As the project grows, so is `__init__.py`. It's maybe time to split it into sub-parts.

It would make clear what is exposed as API and also avoid stuff like `import os as _os`.

t20100:
This PR splits the `__init__.py` into:
- `_filters.py` which provides the filter IDs and configuration classes
- `_utils.py` which provides the filter registration and `get_config` function.

The saving of the build configuration is changed from a `dict` to a `namedtuple` in `setup.py` to be used directly and avoid a circular import.

closes #202


vasole (approved):
LGTM. Tests are unmodified but continue to pass.

PR #206, opened 2022-11-22
Add SZ

vasole:
Include and completes #203


vasole:
At this point, the code compiles under all the platforms and the test file supplied by @orioltinto is properly read.

Still missing tests.

vasole:
Following #207, the default configuration of the SZ filter does not allow it to use it.

Should we provide a default with a relative error?

orioltinto:
SZ is a lossy compressor, the suitability of the proper compression parameters depends very much on the data.
The pw_rel mode stands for pointwise_relative, it means that the values will be within a relative tolerance for each data value. By contrast, the relative mode uses a threshold that is relative to the data range of the full dataset.

To illustrate that, let’s imagine that we have a dataset with values that go from 1 to 1001, if we use the **rel** mode with a threshold of 0.01, the dataset will contain errors that are smaller than (1001-1)*0.01 = 10. If we use the **pw_rel** mode with the same threshold, each point of the dataset will have errors that are smaller than a 1% i.r.t. it's original value. That means that the values in the higher end of the range will have errors smaller than 10.01 and the values in the lower end of the range will have errors smaller than 0.01. Please note that using the rel mode in this example negative values can happen while in the pw_rel mode that's no longer true.

Anyway, I'm not 100% sure about it but I doubt that most data needs more than 6 or 7 decimal values. a default like a pw_rel=10*-6 could be something pretty safe.

vasole:
OK, then the default should be pointwise_relative. Something like 0.0001 % discrepancy with the original data is indeed very conservative. Even too much for my standards.

I would set the default to match the defaults of numpy allclose:

https://numpy.org/doc/stable/reference/generated/numpy.allclose.html

orioltinto:
I was trying to add the docstring directly to the SZ branch in this repository but I don't have permissions. Should I push here directly or to my fork?

vasole:
I am waiting for the review of @t20100 to see if we can already merge this into the main branch.

In the mean time, you can fork this branch and make PR to this branch.

src/hdf5_dl.c
```diff
@@ -449,3 +468,7 @@ herr_t H5Zregister(const void *cls)
 {
 CALL(-1, H5Zregister, cls)
 }
+
+herr_t H5Zunregister(H5Z_filter_t id){
+    CALL(0, H5Zunregister, id)
```
t20100, line 473: Use a negative value for fallback as it is what this function returns in case of error:

```suggestion
    CALL(-1, H5Zunregister, id)
```

src/hdf5_dl.c
```diff
@@ -313,6 +323,10 @@ herr_t H5Epush2(hid_t err_stack, const char *file, const char *func, unsigned li
     }
 }
 
+herr_t H5Eprint2( hid_t err_stack, FILE * stream ){
+    CALL(0, H5Eprint2, err_stack, stream)
```
t20100, line 327: Use a negative value as fallback as it is what this function returns in case of error:
```suggestion
    CALL(-1, H5Eprint2, err_stack, stream)
```

setup.py
```diff
@@ -822,6 +868,58 @@ def prefix(directory, files):
...
+if ZLIB_FROM_BLOSC:
+    sz_sources += glob(blosc_dir + 'internal-complibs/zlib*/*.c')
+    ##depends += glob(blosc_dir + 'internal-complibs/zlib*/*.h')
+else:
+    sz_sources += glob('src/SZ/zlib/*.c')
+
+sz_include_dirs += glob('src/SZ/zstd')
+sz_include_dirs += glob('src/SZ/zstd/common')
+
+if ZLIB_FROM_BLOSC:
```
t20100, line 888: To be consistent with other env var to set the build, let's add a `HDF5PLUGIN_` prefix:
```suggestion
HDF5PLUGIN_ZLIB_FROM_BLOSC = True
if HDF5PLUGIN_ZLIB_FROM_BLOSC:
    sz_sources += glob(blosc_dir + 'internal-complibs/zlib*/*.c')
    ##depends += glob(blosc_dir + 'internal-complibs/zlib*/*.h')
else:
    sz_sources += glob('src/SZ/zlib/*.c')

sz_include_dirs += glob('src/SZ/zstd')
sz_include_dirs += glob('src/SZ/zstd/common')

if HDF5PLUGIN_ZLIB_FROM_BLOSC:
```

setup.py
```diff
@@ -822,6 +868,58 @@ def prefix(directory, files):
...
+
+sz_plugin = HDF5PluginExtension(
+    "hdf5plugin.plugins.libh5sz",
+    sources=sources,
+    depends=depends,
+    include_dirs=include_dirs,
+    extra_compile_args=extra_compile_args,
+    extra_link_args=extra_link_args,
+)
+PLUGIN_LIB_DEPENDENCIES['sz'] = 'sz'
```
t20100, line 921: Not sure this is needed as `sz_sources` is added to `sources`, but we can check that in another PR.

src/hdf5plugin/_filters.py
```diff
@@ -414,6 +418,90 @@ def __init__(self,
...
+
+        self.filter_options = compression_opts
+
+    @staticmethod
+    def pack_error(error: float) -> tuple:
+        from struct import pack, unpack
+        packed = pack('<d', error)  # Pack as IEEE 754 double
+        high = unpack('<I', packed[0:4])[0]  # Unpack high bits as unsigned int
+        low = unpack('<I', packed[4:8])[0]
+        return low, high
```
t20100, line 502: `struct` is already imported at the top level, let's use it:

```suggestion
        packed = struct.pack('<d', error)  # Pack as IEEE 754 double
        high = struct.unpack('<I', packed[0:4])[0]  # Unpack high bits as unsigned int
        low = struct.unpack('<I', packed[4:8])[0]
        return low, high
```

src/hdf5plugin/test.py
```diff
@@ -64,12 +64,18 @@ def _test(self,
         """
         data = numpy.ones((1000,), dtype=dtype).reshape(100, 10)
         filename = os.path.join(self.tempdir, "test_" + filter_name + ".h5")
+        print(filter_name,
+              dtype,
+              lossless,
+              compressed,
+              options)
```
t20100, line 71: Let's remove the print:
```suggestion
```

src/hdf5plugin/_filters.py
```diff
@@ -414,6 +418,90 @@ def __init__(self,
...
+            sz_mode = 0
+            parameter = absolute
+        elif relative is not None:
+            sz_mode = 1
+            parameter = relative
+        elif pointwise_relative is not None:
+            sz_mode = 10
+            parameter = pointwise_relative
+        else:
+            raise NotImplementedError("One of the options need to be provided: abs, rel or pw_rel .")
```
t20100, line 486: Is this meant to be implemented at some point?
Else I propose to use `TypeError` which sounds more appropriate and is used by e.g. `h5py.Group.create_dataset("name")` for missing arguments:

```suggestion
            raise TypeError("One of the options need to be provided: absolute, relative or pointwise_relative.")
```

doc/information.rst
```diff
@@ -56,6 +56,7 @@ HDF5 compression filters and compression libraries sources were obtained from:
 * **FCIDECOMP plugin** (v1.0.2) and **CharLS** (1.x branch, commit `25160a4 <https://github.com/team-charls/charls/tree/25160a42fb62e71e4b0ce081f5cb3f8bb73938b5>`_):
   ftp://ftp.eumetsat.int/pub/OPS/out/test-data/Test-data-for-External-Users/MTG_FCI_Test-Data/FCI_Decompression_Software_V1.0.2 and
   https://github.com/team-charls/charls
+* **SZ plugin** (commit `f36afa4ae9 https://github.com/szcompressor/SZ/commit/f36afa4ae9da0a8c50da3b9e89bfa89c8908b200>`_) and https://github.com/szcompressor/SZ
```
t20100, line 59:
```suggestion
* **SZ plugin** (commit `f36afa4ae9 https://github.com/szcompressor/SZ/commit/f36afa4ae9da0a8c50da3b9e89bfa89c8908b200>`_) and `SZ <https://github.com/szcompressor/SZ>`_
```

t20100 (approved):
LGTM, I left a few suggestions.
There is probably a few things to change in `setup.py` but that can be done in another PR.

doc/information.rst
vasole: SZ plugin is inside SZ.

setup.py
vasole: OK

Issue #204, opened 2022-11-11
Can't write files if using HDF5_PLUGIN_PATH .


orioltinto:
Hello,

because it would be great to use the filters outside python as well, I was trying to use them by setting the environment variable **HDF5_PLUGIN_PATH**.

If I set this variable and I try to read a compressed file without loading hdf5plugin it works without problems (**GOOD!**).

However, when I try to use the filters to write a file it raises an **error from h5py**:

```python
  File "/usr/local/lib/python3.10/dist-packages/h5py/_hl/group.py", line 161, in create_dataset
    dsid = dataset.make_new_dset(group, shape, dtype, data, name, **kwds)
  File "/usr/local/lib/python3.10/dist-packages/h5py/_hl/dataset.py", line 156, in make_new_dset
    dset_id = h5d.create(parent.id, name, tid, sid, dcpl=dcpl, dapl=dapl)
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "h5py/h5d.pyx", line 87, in h5py.h5d.create
ValueError: Unable to create dataset (error during user callback)
```

Also, if the environment variable is set, when importing **hdf5plugin** the libraries are **not initialized** so it **keeps failing**.

In summary, we **need to define** HDF5_PLUGIN_PATH to allow other tools to use the filters but it makes hdf5plugin **fail**.

t20100:
Hi,

Yes you can only read when setting the `hdf5plugin` filters through `HDF5_PLUGIN_PATH`.

The problem is that for writing, some filters need to call some functions of `libhdf5` to get e.g., the data type or the item size, and so you need to link the filter with libhdf5.... This is an issue because the `libhdf5` that one use at runtime through a `h5py` wheel for example is not known at compile time of `hdf5plugin` + C symbols for Python extension are not loaded in the global namespace so not accessible to other packages.

To workaround this `hdf5plugin` provides stubs of `libhdf5` that are initialized at runtime and dynamically load C symbols from the `libhdf5` used by `h5py`.

For the issue of defining  `HDF5_PLUGIN_PATH` making it not loading the filters, it's indeed not good. We never had this use case so far, it was either `HDF5_PLUGIN_PATH` or `h5py`.
[...]

PR #208, opened 2022-11-22
Added `register` function

t20100:
This PR adds a `register(filters=FILTERS, force=True)` function to the API.
With this function it is possible to manually re-register filters embedded in `hdf5plugin`.

The main use case should be to make sure embedded filters are use, which is done with:
```python
import hdf5plugin
hdf5plugin.register()
```

Related to #204

### DIFF
--- a/src/hdf5plugin/_utils.py
+++ b/src/hdf5plugin/_utils.py
@@ -60,7 +60,7 @@ def is_filter_available(name):
     return h5py.h5z.filter_avail(filter_id) > 0
 
 
-REGISTERED_FILTERS = {}
+registered_filters = {}
 """Store hdf5plugin registered filters as a mapping: name: (filename, ctypes.CDLL)"""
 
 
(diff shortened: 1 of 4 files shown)


vasole (commented):
It seems not all versions of h5py wrap the needed library functions

t20100:
> I understand that this doesn't change the behavior when the filter is imported but will enable the possibility to force loading the filters, is that right?

Exactly.

> It seems not all versions of h5py wrap the needed library functions

Yes, h5py< v2.10 does not provide `h5z.unregister_filter` which is needed to re-register a filter.
I added a special case for it.

src/hdf5plugin/_utils.py
```diff
@@ -44,71 +44,139 @@
...
+        elif is_filter_available(name) is True:  # Registered elsewhere
+            filters[name] = "unknown"
+
+    return HDF5PluginConfig(build_config, filters)
+
+
+def register(filters=tuple(FILTERS.keys()), force=True):
+    """Initialise and register `hdf5plugin` embedded filters given their names.
+
+    Unregister corresponding previously registered filters if any.
```
t20100, line 161: Fix docstring
```suggestion
```

t20100:
> I understand that this doesn't change the behavior when the filter is imported but will enable the possibility to force loading the filters, is that right?

This is to be conservative and keep the current behavior.

PR #209, opened 2022-11-22
Add SZ: Docstrings, self-explanatory arguments and default value.

orioltinto:
My changes:
- Docstring with a basic description and a link to the reference webpage.
- Renamed arguments to be self-explanatory (absolute, relative, pointwise_relative)
- Added a default value as pointwise_relative=1.e-05 to match numpy's allclose.


vasole:
One should check if absolute has been set and used, then if relative has been set and used instead of raising an error if two arguments have been set.

The tests are failing because of that.

vasole (approved):
I will correct in the SZ branch.

PR #210, opened 2022-11-23
Min supported version: Python3.7; Depreacted `hdf5plugin.config`

t20100:
~~Merge PR #208 first!~~

This PR sets python 3.7 as the minimal required version (it was 3.4 before) and it makes use of Python3.7 features:
- f-strings (was already used in SZ class).
- module `__getattr__` to deprecate `hdf5plugin.config`

`hdf5plugin.config` is deprecated in favor of `hdf5plugin.get_config()`.

Note that `PLUGINS_PATH` is not deprecated in favor of `PLUGIN_PATH` because this is intended to be used through
```
export HDF5_PLUGIN_PATH=`python -c "import hdf5plugin; print(hdf5plugin.PLUGIN_PATH)"`
```
And a deprecation warning here is not acceptable!


t20100:
Thanks for the fix!

PR #212, opened 2022-11-24
Added `get_filters` function; Fixed `register`

### LINKED ISSUE #207: Adding dictionaries to point to filter classes - not IDs
I am interested in adding GUI support for `hdf5plugin` filters to [NeXpy](https://nexpy.github.io/nexpy/index.html), but it would be useful if there were a way to access both the installed filters _and_ their default options. At the moment, I can find which filters are installed.
```
>>> import hdf5plugin
>>> hdf5plugin.FILTERS
{'blosc': 32001,
 'bshuf': 32008,
 'lz4': 32004,
 'zfp': 32013,
 'zstd': 32015,
 'fcidecomp': 32018}
```
However, I can only find the compression options if I instantiate their respective classes.
```
>>> hdf5plugin.Blosc()['compression']
32001
>>> hdf5plugin.Blosc()['compression_opts']
(0, 0, 0, 0, 5, 1, 1)
```
It would be helpful (to me) if there were an additional dictionary defined in `hdf5plugin.__init__` that linked the filters to the classes, e.g.,
```
INSTALLED_FILTERS = {'blosc': Blosc,
                    'bshuf': Bitshuffle,
                    'bzip2': BZip2,
                    'lz4': LZ4,
                    'zfp': Zfp,
                    'zstd': Zstd,
                    'fcidecomp': FciDecomp
                    }
```
This would allow, e.g.,
```
>>> for f in INSTALLED_FILTERS:
        filter = INSTALLED_FILTERS[f]()
        print(f, filter['compression'], filter['compression_opts'])
blosc 32001 (0, 0, 0, 0, 5, 1, 1)
bshuf 32008 (0, 2)
lz4 32004 (0,)
zfp 32013 ()
zstd 32015 (3,)
fcidecomp 32018 ()
```
I wanted to get feedback before issuing a PR in case alternative solutions are already being considered or there are any objections.

t20100:
This PR adds a `get_filters` function to retrieve filter classes from names or IDs.
By default is returns all filter classes.
It accepts a special argument "registered" and then returns classes of all currently registered filters.

The `register` function was also extended to supports IDs.
The `FILTERS` constant was removed from the documentation (not deprecated though): one can get filter classes with `get_filters` and then retrieve the `filter_id` from the filter classes.

A `filter_name` class attribute was also added to filter classes so one can get it from a class and it is shown in the documentation along with the classes.
Question: Is it really worth adding this `filter_name` which is specific to `hdf5plugin` to the API or not?

@rayosborn would that fits your need?

This PR also fixes a bug in `register` when one filter fails to register, the other are skipped: [58c56eb](https://github.com/silx-kit/hdf5plugin/pull/212/commits/58c56eb77006fd5b69368f0bc44e45071510827d)

closes #207

### DIFF
--- a/doc/contribute.rst
+++ b/doc/contribute.rst
@@ -57,7 +57,7 @@ This briefly describes the steps to add a HDF5 compression filter to the zoo.
 
 * Add a compression options helper class named ``FilterName`` in ``hdf5plugins/_filters.py`` which should inherit from ``_FilterRefClass``.
   This is intended to ease the usage of ``h5py.Group.create_dataset`` ``compression_opts`` argument.
-  It must have a `_filter_name` class attribute with the same name as in the extension defined in ``setup.py`` (without the ``libh5`` prefix) .
+  It must have a `filter_name` class attribute with the same name as in the extension defined in ``setup.py`` (without the ``libh5`` prefix) .
   This name is used to find the filter library.
 
 * Add ``FilterName`` to ``hdf5plugin._filters.FILTER_CLASSES``.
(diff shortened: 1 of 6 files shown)


rayosborn:
Thank you for this. It goes much further than I was proposing and the `registered` option in the `get_fllters` function looks especially useful.

vasole (approved):
Great work! (as usual)

PR #213, opened 2022-11-24
Updated SZ support

t20100:
This PR follows #206 to add and update a few things.

### DIFF
--- a/src/hdf5plugin/_filters.py
+++ b/src/hdf5plugin/_filters.py
@@ -469,18 +469,20 @@ class SZ(_FilterRefClass):
         # Get SZ encoding options
         if absolute is not None:
             sz_mode = 0
-            parameter = absolute
         elif relative is not None:
             sz_mode = 1
-            parameter = relative
         elif pointwise_relative is not None:
             sz_mode = 10
-            parameter = pointwise_relative
         else:
             raise TypeError("One of the options need to be provided: absolute, relative or pointwise_relative.")
 
-        packed_error = self.pack_error(parameter)
-        compression_opts = (sz_mode, *packed_error, *packed_error, *packed_error, *packed_error)
+        compression_opts = (
+            sz_mode,
+            *self.__pack_float64(absolute or 0.),
+            *self.__pack_float64(relative or 0.),
+            *self.__pack_float64(pointwise_relative or 0.),
+            *self.__pack_float64(0.),  # psnr
+        )
 
         logger.info(f"SZ mode {sz_mode} used.")
         logger.info(f"filter options {compression_opts}")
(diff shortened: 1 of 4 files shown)


vasole (approved):
LGTM

src/hdf5plugin/_filters.py
```diff
@@ -466,34 +465,39 @@ class SZ(_FilterRefClass):
...
 
-        packed_error = self.pack_error(parameter)
-        compression_opts = (sz_mode, *packed_error, *packed_error, *packed_error, *packed_error)
+        compression_opts = (
+            sz_mode,
+            *self.__pack_float64(absolute or 0.),
+            *self.__pack_float64(relative or 0.),
+            *self.__pack_float64(pointwise_relative or 0.),
+            *self.__pack_float64(0.),  # psnr
+        )
```
t20100, line 488: Previous implementation was working because not all options are used depending on the mode, but this better matches the meaning of the `compression_opts`

src/hdf5plugin/_filters.py
```diff
@@ -466,34 +465,39 @@ class SZ(_FilterRefClass):
...
-    def pack_error(error: float) -> tuple:
-        packed = struct.pack('<d', error)  # Pack as IEEE 754 double
-        high = struct.unpack('<I', packed[0:4])[0]  # Unpack high bits as unsigned int
-        low = struct.unpack('<I', packed[4:8])[0]
-        return low, high
+    def __pack_float64(error: float) -> tuple:
+        packed = struct.pack('>d', error)  # Pack as big-endian IEEE 754 double
+        high = struct.unpack('>I', packed[0:4])[0]  # Unpack most-significant bits as unsigned int
+        low = struct.unpack('>I', packed[4:8])[0]  # Unpack least-significant bits as unsigned int
+        return high, low
```
t20100, line 500: Same here previous implementation was working but the C code state working with big endian, so let's match it too.

BTW, that would be good to check it on a big-endian machine.

src/hdf5plugin/test.py
```diff
@@ -337,8 +337,7 @@ def test_register_single_filter_by_id(self):
     @unittest.skipUnless(BUILD_CONFIG.embedded_filters, "No embedded filters")
     def test_register_all_filters(self):
         """Re-register embedded filters all at once"""
-        status = hdf5plugin.register()
-        self.assertTrue(status)
+        hdf5plugin.register()
         for filter_name in BUILD_CONFIG.embedded_filters:
```
t20100, line 341: The assert fails if not all filters are embedded (by using `HDF5PLUGIN_STRIP`)

src/hdf5plugin/_filters.py
```diff
@@ -466,34 +465,39 @@ class SZ(_FilterRefClass):
...
             sz_mode = 1
-            parameter = relative
-        elif pointwise_relative is not None:
-            sz_mode = 10
-            parameter = pointwise_relative
         else:
-            raise TypeError("One of the options need to be provided: absolute, relative or pointwise_relative.")
+            sz_mode = 10
+            if pointwise_relative is None:
+                pointwise_relative = 1e-5
```
t20100, line 480: There looks to be a lot more modes:
https://github.com/silx-kit/hdf5plugin/blob/daf5703f744d4ba3d6ab4c9980c99cd4e5c96bcb/src/SZ/sz/include/defines.h#L29-L41

So I propose to change a bit the arguments handling to force only one argument for now to prepare for adding support of additional modes.
We can later add for example `ABS_AND_REL` when both `absolute` and `relative` are provided.
But I expect an extra `mode` argument will be needed for the `OR` modes (e.g., `ABS_OR_REL`).

t20100:
attn @orioltinto

src/hdf5plugin/_filters.py
vasole: I would still leave a working default as in the previous code. All our filters are working without modification.

t20100: I kept the same default as before:
```python
            if pointwise_relative is None:
                pointwise_relative = 1e-5
```

vasole: OK. I misunderstood the assert.

PR #215, opened 2022-11-25
Updated SZ build

t20100:
This PR follows #206 and rework the build:
- Use a static `config.h` file, actually only `HAVE_UNISTD_H` is needed.
- Use `zlib` and `zstd` from `c-blosc` rather than included in `SZ`: `zlib` is the same version but for an include of `<unistd.h>`, and `zstd` in `c-blosc` is a more recent (compatible: 1.x) version. This paves the way for building compression libraries only once for all plugins.
- Do not build sz compression as a lib because it is already built with the plugin.


src/SZ/config.h
```diff
@@ -0,0 +1,4 @@
+/* Static config.h file added for hdf5plugin to spare running the configuration */
+#if !defined(_WIN32) || defined(__MINGW32__)
+#define HAVE_UNISTD_H
+#endif
```
t20100, line 4: I was wondering whether to put it in `src/SZ/` or in a separate folder e.g., `src/SZ_extra/`

setup.py
```diff
@@ -872,39 +831,26 @@ def prefix(directory, files):
...
-# "cflags": ['-DBUILD_HDF5_FILTER:BOOL=ON'],
+# zlib
+sz_sources += zlib_sources
+sz_include_dirs += zlib_include_dirs
+sz_depends += zlib_depends
 
+# zstd
+sz_sources += zstd_sources
+sz_include_dirs += zstd_include_dirs
+sz_depends += zstd_depends
```
t20100, line 846: Using sources from `c-blosc` third parties

t20100:
attn @orioltinto

src/SZ/config.h
vasole: I guess in a separate folder is better in order not to mix our code with theirs.

t20100: OK, done!

PR #216, opened 2022-11-25
[WIP] SZ3

vasole:
Preliminary addition


vasole:
MacOS issues:

- It needs a dummy `omp.h`
- It requires `--std=C++14` because of the use of `std::make_unique`

PR #223, opened 2022-11-28
Fixed documentation

t20100:
This PR updates lists of filters in the documentation and the copyright year.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -4,7 +4,7 @@ hdf5plugin
 .. image:: https://zenodo.org/badge/DOI/10.5281/zenodo.7257761.svg
    :target: https://doi.org/10.5281/zenodo.7257761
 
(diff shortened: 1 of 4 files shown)


vasole (commented):
FCIDECOMP is all capitals in https://portal.hdfgroup.org/display/support/Registered+Filter+Plugins

t20100:
I wanted to use the name of the helper classes (I need to fix `BitShuffle`) when listing the filters, and I also wanted to normalize the use of capitals for acronyms: 2 letters long -> capital (e.g. `SZ`) >=3 letters long: Camel case (e.g. `Zfp`)... But here the names of the classes are not following this: `Bitshuffle` instead of `BitShuffle` and `Zstd` instead of `ZStd`.

Note: FCIDECOMP apparently stands for "**F**lexible **C**ombined **I**mager **Decomp**ression".

The use of capitals in https://portal.hdfgroup.org/display/support/Registered+Filter+Plugins is not consistent with the one in the in the page providing the filters' details: https://portal.hdfgroup.org/display/support/Filters .

PR #224, opened 2022-11-29
Starting point to include SZ3

vasole:
This PR provides a starting point to include the filter.

Not all features are tested, but they are exposed to the user.


vasole:
The SZ repository (not the SZ3) contains example test data in text and binary format. The test code inside the H5Z-SZ3 repository is expected to be used (once built and with the dlls in the used HDF5 library repository) as follows:

```
"C:\Program Files\HDF_Group\HDF5\1.10.6\bin\convertBinToHDF5.exe" -f data .\testdata\x86\testfloat_8_8_128.dat 8 8 128
"C:\Program Files\HDF_Group\HDF5\1.10.6\bin\convertBinToHDF5.exe" -d data .\testdata\x86\testdouble_8_8_128.dat 8 8 128

REM	printf("Usage: print_h5repack_args <options>\n");
REM	printf("Options:\n");
REM	printf("	-M <error bound mode> : 10 options as follows. \n");
REM	printf("		ABS (absolute error bound)\n");
REM	printf("		REL (value range based error bound, so a.k.a., VR_REL)\n");
REM	printf("		PSNR (peak signal-to-noise ratio)\n");
REM	printf("		NORM2 (norm2)\n");
REM	printf("	-A <absolute error bound>: specifying absolute error bound\n");
REM	printf("	-R <value_range based relative error bound>: specifying relative error bound\n");
REM	printf("	-N <norm2>: specifying norm2 error bound\n");
REM	printf("	-S <PSNR>: specifying PSNR\n");
REM	printf("* examples: \n");
REM	printf("	print_h5repack_args -M ABS -A 1E-3 (output: -f UD=32024,0,9,0,1062232653,3539053052,0,0,0,0,0,0)\n");
REM	printf("	print_h5repack_args -M REL -R 1E-4 (output: -f UD=32024,0,9,1,0,0,1058682594,3944497965,0,0,0,0)\n");
REM	exit(0);

"C:\Program Files\HDF_Group\HDF5\1.10.6\bin\h5repack.exe" -f UD=32024,0,9,0,1062232653,3539053052,0,0,0,0,0,0 .\testdata\x86\testfloat_8_8_128.dat.h5 .\testdata\x86\testfloat_8_8_128.dat.abs.h5
"C:\Program Files\HDF_Group\HDF5\1.10.6\bin\h5repack.exe" -f UD=32024,0,9,1,0,0,1058682594,3944497965,0,0,0,0 .\testdata\x86\testfloat_8_8_128.dat.h5 .\testdata\x86\testfloat_8_8_128.dat.rel.h5
"C:\Program Files\HDF_Group\HDF5\1.10.6\bin\h5repack.exe" -f UD=32024,0,9,0,1062232653,3539053052,0,0,0,0,0,0 .\testdata\x86\testdouble_8_8_128.dat.h5 .\testdata\x86\testdouble_8_8_128.dat.abs.h5
"C:\Program Files\HDF_Group\HDF5\1.10.6\bin\h5repack.exe" -f UD=32024,0,9,1,0,0,1058682594,3944497965,0,0,0,0 .\testdata\x86\testdouble_8_8_128.dat.h5 .\testdata\x86\testdouble_8_8_128.dat.rel.h5
```

src/hdf5plugin/_filters.py
```diff
@@ -500,6 +503,75 @@ def __pack_float64(error: float) -> tuple:
...
+    }
+
+
+    For more details about the compressor `SZ <https://szcompressor.org/>`_.
+    """
+    filter_name = "sz3"
+    filter_id = SZ3_ID
+
+    def __init__(self, absolute=None, relative=None, norm2=None, peak_signal_to_noise_ratio=None):
+        if (absolute, relative, norm2, peak_signal_to_noise_ratio, peak_signal_to_noise_ratio).count(None) < 2:
```
t20100, line 535:
```suggestion
        if (absolute, relative, norm2, peak_signal_to_noise_ratio).count(None) < 3:
```

src/hdf5plugin/_filters.py
```diff
@@ -500,6 +503,75 @@ def __pack_float64(error: float) -> tuple:
...
+            exit(0);
+    }
+
+
+    For more details about the compressor `SZ <https://szcompressor.org/>`_.
+    """
+    filter_name = "sz3"
+    filter_id = SZ3_ID
+
+    def __init__(self, absolute=None, relative=None, norm2=None, peak_signal_to_noise_ratio=None):
```
t20100, line 534: For a first version with limited features, I would limit the arguments:
```suggestion
    def __init__(self, absolute=None):
        relative = None
        norm2 = None
        peak_signal_to_noise_ratio = None
```

src/hdf5plugin/_filters.py
```diff
@@ -500,6 +503,75 @@ def __pack_float64(error: float) -> tuple:
...
+            printf("	-S <PSNR>: specifying PSNR\n");
+            printf("* examples: \n");
+            printf("	print_h5repack_args -M ABS -A 1E-3 (output: -f UD=32024,0,9,0,1062232653,3539053052,0,0,0,0,0,0)\n");
+            printf("	print_h5repack_args -M REL -R 1E-4 (output: -f UD=32024,0,9,1,0,0,1058682594,3944497965,0,0,0,0)\n");
+            exit(0);
+    }
+
+
+    For more details about the compressor `SZ <https://szcompressor.org/>`_.
+    """
```
t20100, line 530:
```suggestion
    """``h5py.Group.create_dataset``'s compression arguments for using SZ3 filter.

    It can be passed as keyword arguments:

    - **Absolute** mode: To use, set the ``absolute`` argument.
      It ensures that the resulting values will be within the provided absolute tolerance.
      .. code-block:: python
          f.create_dataset(
              'sz3_absolute',
              data=numpy.random.random(100),
              **hdf5plugin.SZ3(absolute=0.1))

    For more details about the compressor, see `SZ3 <https://szcompressor.org/>`_.
    """
```

src/hdf5/include/H5Epublic.h
```diff
@@ -52,8 +52,14 @@ typedef struct H5E_error2_t {
...
 
 /* HDF5 error class */
+#ifdef __cplusplus
+extern "C" {
+#endif
 #define H5E_ERR_CLS		(H5OPEN H5E_ERR_CLS_g)
 H5_DLLVAR hid_t H5E_ERR_CLS_g;
+#ifdef __cplusplus
+}
+#endif
```
t20100, line 62: Issue here is that it won't build with external headers, is there another way to handle this without a patch

setup.py
```diff
@@ -869,6 +875,148 @@ def prefix(directory, files):
...
+    f.write("#define SZ3_VERSION_HPP\n")
+    f.write("\n")
+    f.write('#define SZ3_NAME "SZ3"\n')
+    f.write('#define SZ3_VER  "3.1.7"\n')
+    f.write("#define SZ3_VER_MAJOR 3\n")
+    f.write("#define SZ3_VER_MINOR 1\n")
+    f.write("#define SZ3_VER_PATCH 7\n")
+    f.write("#define SZ3_VER_TWEAK \n")
+    f.write("\n")
+    f.write("#endif //SZ3_VERSION_HP\n")
```
t20100, line 896: Can't this file be put in `SZ3_extra/`?

setup.py
```diff
@@ -869,6 +875,148 @@ def prefix(directory, files):
...
+                    os.path.join(sz3_dir, "include", "SZ3", "compressor"),
+                    os.path.join(sz3_dir, "include", "SZ3", "encoder"),
+                    os.path.join(sz3_dir, "include", "SZ3", "frontend"),
+                    os.path.join(sz3_dir, "include", "SZ3", "lossless"),
+                    os.path.join(sz3_dir, "include", "SZ3", "predictor"),
+                    os.path.join(sz3_dir, "include", "SZ3", "preprocessor"),
+                    os.path.join(sz3_dir, "include", "SZ3", "quantizer"),
+                    os.path.join(sz3_dir, "include", "SZ3", "utils"),
+                    os.path.join(sz3_dir, "include", "SZ3", "utils", "inih"),
+                    os.path.join(sz3_dir, "include", "SZ3", "utils", "ska_hash"),
```
t20100, line 912: And added to include_dirs:
```suggestion
                    os.path.join(sz3_dir, "include", "SZ3", "utils", "ska_hash"),
                    os.path.join("src", "SZ3_extra"),
```

setup.py
```diff
@@ -869,6 +875,148 @@ def prefix(directory, files):
...
+        './dictBuilder/cover.c',
+        './dictBuilder/divsufsort.c',
+        './dictBuilder/zdict.c',
+        './dictBuilder/fastcover.c',
+        ]
+
+    zstd_extra_objects = []
+    zstd_define_macros = []
+    zstd_sources = []
+    zstd_depends = []
```
t20100, line 974: That's risky if we add more filters below.
I reckon we should improve the reuse of zstd, but in the mean time, what about doing it the same way as for the other filters? : defining `extra_objects`, `define_macros`,...; appending to those and using them in `HDF5PluginExtension`

setup.py
```diff
@@ -869,6 +875,148 @@ def prefix(directory, files):
...
+
+# I have issues with the BMI on linux and Mac because of compiling Zstd separately
+
+if sys.platform.startswith('darwin'):
+    HDF5PLUGIN_ZSTD_FROM_BLOSC = False
+else:
+    HDF5PLUGIN_ZSTD_FROM_BLOSC = True
+
+if not HDF5PLUGIN_ZSTD_FROM_BLOSC:
+    sz3_zstd_cmake_list = [
```
t20100, line 928: May simpler to use `glob` here as for zstd for blosc.

setup.py
```diff
@@ -869,6 +875,148 @@ def prefix(directory, files):
...
+    zstd_include_dirs = [os.path.join(sz3_dir, "tools", "zstd")]
+    sz3_zstd_sources = []
+    for src in sz3_zstd_cmake_list:
+        items = src.split('/')
+        sz3_zstd_sources.append(os.path.join(sz3_dir, "tools", "zstd", items[1], items[2]))
+        zstd_include_dirs.append(os.path.join(sz3_dir, "tools", "zstd", items[1]))
+else:
+    sz3_zstd_sources = zstd_sources
+
+#TODO: Probably OpenMP should not be the default for this filter (sz3.config)
```
t20100, line 984: What is the problem with OpenMP?

setup.py
```diff
@@ -869,6 +875,148 @@ def prefix(directory, files):
...
+    cpp11_required=True,
+    )
+
+if sys.platform.startswith('darwin'):
+    # this should be taken from the output of HDF5PluginExtension
+    sz3_zstd_sources += [os.path.join('src', 'hdf5_dl.c')]
+    zstd_include_dirs += [os.path.join('src', 'hdf5', 'include'), os.path.join('src', 'hdf5', 'include', 'darwin')]
+
+sz3_lib = ("sz3", {
+    "sources": sz3_zstd_sources,
```
t20100, line 1010: This looks to compile zstd twice since it is also add to `sz3_plugin` `sources`.

Actually, this way of adding our patch gives me an idea: do the same for all filters. Instead of adding `hdf5_dl.c`, we could add a static library with it inside

setup.py
```diff
@@ -496,7 +496,13 @@ def __init__(self, name, sse2=None, avx2=None, cpp11=None, cpp11_required=False,
...
 
         else:
-            self.sources.append(os.path.join('src', 'hdf5_dl.c'))
+            if name.endswith("h5sz3") and sys.platform.startswith('darwin'):
+                # MacOS does not like to mix C and C++ code and next line
+                # does not work for the macro CALL
+                #self.sources.append(os.path.join('src', 'hdf5_dl.cpp'))
+                pass
+            else:
+                self.sources.append(os.path.join('src', 'hdf5_dl.c'))
```
t20100, line 505: It should be possible to replace this with adding a static lib (see other comment) and that would remove the need for a specific handling for sz3 in the generic class.

t20100 (commented):
Nice!
Some first comments/suggestions

src/hdf5plugin/_filters.py
vasole: I would not limit the arguments, but only document those used. In any case I am going to try to see how many are working.

src/hdf5/include/H5Epublic.h
vasole: It will not build only under windows. Under linux and Mac the patch is not needed.

setup.py
vasole: I had already tried, however the code explicitly import from "SZ3/version.hpp".

I can try to add a repository SZ3 containing version.hpp inside SZ3_extra to see if it works.

setup.py
vasole: No, because when that is done, the sz3_plugin sources for zstd are empty.

setup.py
vasole: I cut and pasted from their CMakeLists file for the shipped Zstd code.

PR #227, opened 2022-11-29
Needed when compiling as C++ inside extern "C" {}

vasole:
@t20100 

Our default for H5Tget_order does not seem to be correct. It should be negative in case of error.


t20100 (approved):
LGTM

PR #228, opened 2022-11-29
Provide correct HDF5 defaults

vasole:
H5T_ORDER_ERROR is not 0 but negative.

Using HDF5 library constants also solves issues when compiling the code with C++ compilers.


t20100 (approved):
LGTM

PR #234, opened 2022-12-04
Skip zstd test in inner iteration

spanezz:
There is currently no zstd filter plugin in Debian. The test suite of python-hdf5plugin knows how to skip zstd tests accordingly, except for one inner iteration, which needs this extra check

### DIFF
--- a/src/hdf5plugin/test.py
+++ b/src/hdf5plugin/test.py
@@ -143,6 +143,8 @@ class TestHDF5PluginRW(BaseTestHDF5PluginRW):
 
         # Specify options
         for cname in ('none', 'lz4', 'zstd'):
+            if cname != "none" and not should_test(cname):
+                continue
             for dtype in (numpy.int8, numpy.int16, numpy.int32, numpy.int64):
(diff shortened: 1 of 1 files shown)


t20100:
Hi,

The issue is not about the `zstd` filter plugin but about the version of the `bitshuffle` filter plugin: It differs between Debian and the one embedded in `hdf5plugin`.

Debian currently provides v3.5.1 which uses only the `LZ4` compression library while `hdf5plugin` embeds v0.4.2 which adds the option to use either `LZ4` or the `ZStd` compression library, thus the failing test.

The only way I see in `hdf5plugin` to check the version is to write a compressed dataset and check for its flags which contains the version of the filter (I'll prepare a PR for this).

t20100:
I'm closing this as PR  #235 solves the issue by supporting older version of the `bitshuffle` filter in the test.

t20100:
FYI, we tagged a patch release [4.0.1](https://github.com/silx-kit/hdf5plugin/releases/tag/v4.0.1) including PR #235 that should fix the related issue.

PR #235, opened 2022-12-05
Added support of bitshuffle <v0.4 in tests

t20100:
This PR adds a test of the version of the used `bitshuffle` filter to toggle the test of `zstd` compressor option.
This allows to support older versions of the `bitshuffle` filter than the one embedded in `hdf5plugin`.

Related to #234

attn @spanezz


vasole (commented):
Proposed to use in memory file

t20100:
Yes, I hesitated to do so, I'm changing it.

PR #240, opened 2022-12-07
Updated build setup to share some build steps

### LINKED ISSUE #220: Build compression libraries as C lib once
As it is now, compression libraries like `zstd` or `zlib` are build multiple time from source to be used in different filters.
That would be better to build those once as a static lib and link it with the different filters.

t20100:
This PR aims at:
- making `setup.py` hopefully a bit easier to read by splitting the build config of static libs and plugins by using functions rather.
- building static library of compressors shared by multiple filters as well as the `hdf5_dl.c` file to avoid building multiple times the same code.

It also fixes some `flake8` linting issues, in particular a typo in `_utils.py` using `_logger` instead of `logger`.

closes #220

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -595,6 +595,22 @@ def get_snappy_clib(field=None):
     return config[field]
 
 
+def get_zfp_clib():
+    """ZFP static lib build config"""
+    cflags = ['-O3', '-ffast-math', '-std=c99']  # TODO, '-fopenmp']
+    cflags += ['/Ox', '/fp:fast', '/openmp']
+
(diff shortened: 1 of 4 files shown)


t20100:
Ready for review.

Shared compression libs (zlib, lz4, zstd) are built as static libs and so reused by filter plugins.

I also took the opportunity to use python>=3.7 features: f-strings ans `sys.platform` stopping to have the version number as suffix.

vasole (approved):
Hard to review. It looks good to the CI

PR #241, opened 2022-12-07
Add sz3 file for testing

vasole:
- absolute and norm2 modes seem quite reliable
- relative modes seem to give different results following the platform and or compielr


t20100 (approved):
LGTM

PR #246, opened 2022-12-14
Added SZ3 to documentation

### LINKED ISSUE #225: Document SZ3
It is a C++14 filter but the filter can be built in all our tested platforms. See #224

t20100:
Can more mode be added to the doc?

closes #225


vasole:
The information can be found in:

https://github.com/silx-kit/hdf5plugin/blob/main/src/SZ3/tools/sz3/sz3.config

We can consider adding the norm2 documentation:

`norm2 = sqrt((x1-x1')^2+(x2-x2')^2+....+(xN-xN')^2)`

PR #250, opened 2023-01-06
Updated c-blosc (v1.21.2) and lz4 compression lib (1.9.4)

t20100:
This PR:
- updates c-blosc to v1.21.2
- changes the used `lz4` and `ztsd` compression libraries to those embedded in c-blosc2 rather than c-blosc. This updates the version of `lz4` used to v1.9.4 as raised in https://github.com/silx-kit/hdf5plugin/issues/144#issuecomment-1354801581 by @jonwright
- It also imports `CompileError` from `setuptools` when available as a small follow-up of PR #248


vasole:
The test with oldest h5py version will fail if combined with newest numpy.

It seems numpy.typeDict has been deprecated for quite long time and numpy.sctypeDict should be used.

vasole:
I guess we should set NUMPY_OLDER_VERSION for that particular python 3.8 matrix.

vasole (changes requested):
Set NUMPY_OLDER_VERSION to 1.21.0 (for instance) in the failing matrix.

t20100:
I set `NUMPY_OLDER_VERSION` for the failing test

vasole (approved):

PR #256, opened 2023-01-16
Updated documentation

t20100:
This PR reworks the list of filter and compression libs source links.
The goal is to make it simpler to maintain by avoiding duplicated links to compression libs using by mulitple filters.
It also simplifies the display by not displaying the URLs any more.

Screenshot from 2023-01-16 10-31-46

See previous: http://www.silx.org/doc/hdf5plugin/4.0.0/information.html#hdf5-filters-and-compression-libraries


t20100:
Replace reference to a commit in PyTables with proper release.

vasole (approved):

Issue #254, opened 2023-01-11
Add compilation of blosc2 "plugins"


t20100:
Following addition of blosc2 (PR #201).

As it is, code in `c-blosc2/plugins` is not built.
Those plugins cannot be used for writing through the DHF5 blosc2 filter, but building those would allow to read chunks written directly with the `blosc2` library.

PR #252, opened 2023-01-06
Added Intel IPP optional support for LZ4 compression; Added `HDF5PLUGIN_INTEL_IPP_DIR` configuration environment variable

t20100:
~~Merge PR #251 first!~~

This PR adds optional support of Intel IPP libraries for:
- The LZ4 compression library: Using the patch provided with IPP.
- Replacement of LZ4 calls in c-blosc2 (through `HAVE_IPP` macro)

It adds a `HDF5PLUGIN_IPP_DIR` env. var. to configure IPP root dir.

closes #144

PR #257, opened 2023-01-16
Updated build configuration

t20100:
This PR partly reverts PR #255 so that it ensures SSE2/AVX2 support is disabled in blosc/blosc2 when requested (else it could be enabled with HDF5PLUGIN_NATIVE=True...).
This also reworks the way to handle SSE2 intrinsics support on P9: It makes it explicit in bitshuffle and use altivec for blosc2 rather than combining it with SSE2: Now SSE2 is only for x86.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -885,7 +885,10 @@ def get_blosc2_plugin():
     extra_link_args = []
     libraries = []
 
-    sse2_kwargs = {'define_macros': [('SHUFFLE_SSE2_ENABLED', 1)]}
+    if platform.machine() == 'ppc64le':
+        sse2_kwargs = {}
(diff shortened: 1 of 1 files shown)


t20100:
Added a fix for link issue building with IPP that was added in #254 where the IPP libs were not loaded for lz4 and bitshuffle filters.
First raised here: https://github.com/silx-kit/hdf5plugin/pull/252#issuecomment-1378482110

With gcc on ubuntu20.04 (gcc9) (not on debian10 from what I tested gcc8), order of libraries matter, so this PR adds `lz4` **before** IPP libs.

vasole (approved):

PR #268, opened 2023-06-16
Fixed SZ compression filter build options

### LINKED ISSUE #267: SZ compressor with absolute mode.
I'm having a problem with the SZ compressor with the absolute mode.
I [opened an issue](https://github.com/szcompressor/SZ/issues/108) at the SZ github page but after digging a lot my best guess is that the problem must be related on how the plugin is built in hdf5plugins.

A script to reproduce the error:

```python
import tempfile

import h5py
import hdf5plugin
import numpy as np

# Some parameters
SHAPE = (1000, 25, 25)
TOLERANCE = 0.01


# Generate random data
np.random.seed(0)
data = np.random.random(size=SHAPE).astype(np.float32)

encoding = hdf5plugin.SZ(absolute=TOLERANCE)
# Add chunking info
encoding = {**encoding, "chunks": data.shape}


with tempfile.NamedTemporaryFile() as tmp_file:
    # Create compressed file
    with h5py.File(tmp_file.name, 'w') as f:
        f.create_dataset('var', data=data, **encoding)

    # Open compressed file
    with h5py.File(tmp_file.name, 'r') as f:
        recovered_data = f["var"][:]

# Check that the data fulfills the constrain
if not np.allclose(data, recovered_data, atol=TOLERANCE):
    max_diff = np.max(np.abs(recovered_data - data))
    raise AssertionError(f"Condition not fulfilled for {TOLERANCE=} -> {max_diff=}")
```

The error can be reproduced in a Docker container with python:3.11 . I also encountered the problem with older versions.

```Docker
FROM python:3.11
RUN pip install hdf5plugin
```

t20100:
This PR changes SZ compression filter build options to fix an issue with SZ absolute mode.
I added the test from issue #267
  
closes #267

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1081,8 +1081,8 @@ def get_sz_plugin():
     sz_dir = "src/SZ/sz"
     h5zsz_dir = "src/SZ/hdf5-filter/H5Z-SZ"
 
-    extra_compile_args = ['-O3', '-ffast-math', '-std=c99', '-fopenmp']
-    extra_compile_args += ['/Ox', '/fp:fast', '/openmp']
+    extra_compile_args = ['-O3', '-std=c99', '-fopenmp']
+    extra_compile_args += ['/Ox', '/openmp']
     extra_link_args = ['-fopenmp', '/openmp', "-lm"]
 
     include_dirs = [f'{h5zsz_dir}/include']
--- a/src/hdf5plugin/test.py
+++ b/src/hdf5plugin/test.py
@@ -444,18 +444,16 @@ class TestSZ(unittest.TestCase):
 
         compression = hdf5plugin.SZ(absolute=tolerance)
 
-        with tempfile.NamedTemporaryFile() as tmp_file:
-            # Create compressed file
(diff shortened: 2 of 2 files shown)


vasole:
There is a problem with line 458 of test.py

vasole:
image

vasole:
Why not use the core driver therefore avoiding the need to use bytes io?

vasole:
We have already used it:

https://github.com/silx-kit/hdf5plugin/blob/e01716c6e4cf9e1590edec2e8e1453e9458497c8/test/test.py#LL218C20-L218C20

t20100:
Indeed, didn't though of the 'core' driver. BTW `BytesIO` approach does not work with old `h5py`

t20100:
Tests should pass now

robertu94:
When will this be available in the `pip install`-able copy?  Should it just be there?

vasole:
> When will this be available in the `pip install`-able copy? Should it just be there?

It is already the case :)

robertu94:
Awesome!  Thanks for your help on this :)

PR #271, opened 2023-09-04
Updated documentation

t20100:
Small update to fix layout of list in sphinx documentation and update readthedoc configuration that's using an obsolete parameter

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -25,4 +25,3 @@ python:
      - requirements: requirements.txt
      - method: pip
        path:  .
-   system_packages: true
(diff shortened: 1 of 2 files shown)


t20100:
CI fails on macos with Python3.7 because it now produces a `macosx_11_7_x86_64.whl` wheel while in produced a `macosx_10_15_x86_64.whl` last time it run....

`ERROR: hdf5plugin-4.1.3-py3-none-macosx_11_7_x86_64.whl is not a supported wheel on this platform` 

Something has changed and `MACOSX_DEPLOYMENT_TARGET` no longer seems to be taken into account:
`wheel` and `setuptools` has been updated in the meantime (and `setuptools` dropped python3.7 along the way), also macOS version changed from 12.6.5 to 12.6.7...

kif:
RTD is not part of the CI ... it should not be too complicated to integrate it...

t20100:
RTD is set-up through a github webhook

kif (approved):
LGTM

PR #272, opened 2023-09-04
Updated dependency to `wheel`>=0.34.0

t20100:
This PR requires at least wheel>=[0.34.0](https://pypi.org/project/wheel/0.34.0/) (from January 2020).
This simplifies a bit `setup.py` and remove deprecated dependency on `pkg_resources`


.github/workflows/ci.yml
```diff
@@ -37,7 +37,7 @@ jobs:
             H5PY_OLDER_VERSION: 2.10.0
             NUMPY_OLDER_VERSION: 1.23.5
 
-          - name-suffix: "sdist-h5py_3.8.0"
```
t20100, line 40: It was not aligned with the task config

kif (approved):
Simplification is always good.

Issue #193, opened 2022-10-21
Remove dependency on `distutils`


t20100:
`hdf5plugin` packaging relies on `setuptools` and `distutils`, the later being deprecated.

As of today, some of the feature used from `distutils` are not available from `setuptools`, for instance: `from distutils import ccompiler, error`

It might be worth checking if `meson` suits our needs.

t20100:
[cmake-python-distributions](https://github.com/scikit-build/cmake-python-distributions) is also worth checking since some of the embedded projects are using cmake.

vasole:
setuptools is in fact including its own copy of distutils

As far as I have understood, if you import setuptools before importing anything from  distutils, the internal copy of setuptools is used.

eli-schwartz:
> setuptools is in fact including its own copy of distutils
> 
> As far as I have understood, if you import setuptools before importing anything from distutils, the internal copy of setuptools is used.

That's not a solution -- in fact, it's part of the problem.

If you're relying on distutils APIs, then even on python 3.11, which *has* distutils in the stdlib, setuptools installs an injection hook that makes `python -c "import distutils; print(distutils)"` show that distutils is `..../site-packages/setuptools/_distutils/__init__.py`

And setuptools does NOT want you to import from distutils. Doing so is deprecated, and they are developing their vendored distutils with the goal of fixing issues in it by... changing the API. Or deprecating things they don't think people are using, even if they turn out to be using it.

t20100:
I 'm closing this issue: `main` only imports `distutils` as a fallback for old versions of `setuptools` (<62.4).

PR #276, opened 2023-09-07
Removed import of `distutils`

t20100:
This PR aims at removing the import of `distutils` by using an approach proposed in https://github.com/pypa/setuptools/issues/2806#issuecomment-961805789.

Imports of `distutils` as fallbacks for older versions of `setuptools` are kept for now.

related to  #193


t20100:
Added `wheel` dependency to debian packaging that was added as requirement in PR #272

kif (approved):
LGTM

PR #282, opened 2023-10-23
Update hdf5-blosc2 to support b2nd

ivilata:
This provides a few updates to support Blosc2 NDim NDim (2-level partitioning or b2nd, as introduced in Blosc2 2.7.0, see <https://www.blosc.org/posts/blosc2-ndim-intro/>) for arrays with rank >= 2. The code comes from PyTables master commit bb02c88 (where it passes all tests); part of it was already released in PyTables v3.9.0 and the rest will be in the next version (though it is completely compatible with the released one and not expected to change until then). A minor documentation fix and an updated unit test are also included.

For such b2nd-enabled arrays, each HDF5 chunk is stored as a Blosc2 multidimensional array with a fixed shape (even for margin chunks where data may not cover the whole chunk), consisting of a single inner chunk itself, and it also adds the HDF5 chunk rank and shape to stored filter parameters (so that further chunks added with the filter have consistent chunk and block sizes).

To enable reading datasets generated by other code, the filter is more lax regarding the inner structure of Blosc2 arrays (they may consist of several inner chunks or use different block sizes).  It may also work without the extra filter parameters (though it has to skip some data validation, so it warns about this).

See PyTables/PyTables#1056 and PyTables/PyTables#1072 for more information.

### DIFF
--- a/doc/information.rst
+++ b/doc/information.rst
@@ -57,7 +57,7 @@ HDF5 compression filters and compression libraries sources were obtained from:
   using `BZip2 <https://sourceware.org/git/bzip2.git>`_ (v1.0.8).
 * `hdf5-blosc plugin <https://github.com/Blosc/hdf5-blosc>`_ (v1.0.0)
   using `c-blosc <https://github.com/Blosc/c-blosc>`_ (v1.21.5), LZ4, Snappy, ZLib and ZStd.
-* hdf5-blosc2 plugin (from `PyTables <https://github.com/PyTables/PyTables/>`_ v3.9.2.dev0)
+* hdf5-blosc2 plugin (from `PyTables <https://github.com/PyTables/PyTables/>`_ v3.9.2.dev0, commit `3ba4e78 <https://github.com/PyTables/PyTables/tree/3ba4e78336f21f5e32ddf49ced4560f610de70dd/hdf5-blosc2/src>`_)
   using `c-blosc2 <https://github.com/Blosc/c-blosc2>`_ (v2.10.2), LZ4, ZLib and ZStd.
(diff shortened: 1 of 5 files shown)


t20100:
Thanks for the PR!

Could you also update the information about which version or commit of `PyTables` the filter comes from:
https://github.com/silx-kit/hdf5plugin/blob/bd82712ba6146f6f973663c64262e81bf77bfd94/doc/information.rst?plain=1#L60

We try to keep the list of versions embedded in `hdf5plugin` up-to-date.

ivilata:
Sure! Updated to PyTables v3.9.2.dev0 in commit 7831797d.

t20100 (approved):
LGTM, thanks again for the PR!

ivilata:
Thanks @t20100! I was going to add that I'll be doing some extra testing (via PyTables) during the next days to check for more cases. Shall I fix anything, I'll ping you ASAP. I hope to be able to complete this testing before a new hdf5plugin release happens!

t20100:
Hi,

> Would it be ok with you if a new PR updated the version of C-Blosc2?

We try to keep up-to-date with upstream projects, so a PR with an update to the next release of c-blosc2 is welcomed!

> I could post that PR first, then another one to fix hdf5-blosc2 code.

Sounds good to me.

ivilata:
Hi @t20100, I found out that the problem is contained in PyTables-specific code for optimized array slicing, and filter code is actually not affected, so the code in this PR should be ok as is AFAIK, and hdf5plugins doesn't need to update C-Blosc2 for it to work.

Thanks again, and sorry for the noise!

t20100:
Thanks for the information!
I'm anyway updating to latest release of `c-blosc2` (#283) to enable AVX512.

PR #283, opened 2023-11-06
Updated `c-blosc2` to v2.11.1; Enabled AVX512 support in `c-blosc2`

t20100:
This PR updates `c-blosc2` to v2.11.1 which adds AVX512 bitshuffle implementation.

This still need some update/test to enable AVX512 support.


t20100:
Ready for review.

This compiles blosc2 with AVX512 support if enabled: The `SHUFFLE_AVX512_ENABLE` is always set and `bitshuffle-avx512.c` always compiled and the selection of whether or not AVX512 support is available relies on AVX512 compiler arguments (`-mavx512f` `-mavx512bw` `/arch:AVX512`) and  `#if defined(__AVX512F__) && defined (__AVX512BW__)` in `bitshuffle-avx512.c`.

PR #289, opened 2024-01-19
Updated embedded libraries: SZ3 v3.1.8

t20100:
This PR aims at upgrading SZ3 to the latest release: https://github.com/szcompressor/SZ3/releases/tag/v3.1.8

### DIFF
--- a/doc/information.rst
+++ b/doc/information.rst
@@ -83,8 +83,7 @@ HDF5 compression filters and compression libraries sources were obtained from:
 * `SZ plugin <https://github.com/szcompressor/SZ>`_
   (commit `c25805c12b3 <https://github.com/szcompressor/SZ/commit/c25805c12b339d2cb2f406f95293b9a7313c4fb1>`_)
   using `SZ <https://github.com/szcompressor/SZ>`_, ZLib and ZStd.
-* `SZ3 plugin <https://github.com/szcompressor/SZ3>`_
-  (commit `4bbe9df7e4bcb <https://github.com/szcompressor/SZ3/commit/4bbe9df7e4bcb6ae6339fcb3033100da07fe7434>`_)
+* `SZ3 plugin <https://github.com/szcompressor/SZ3>`_ (v3.1.8)
   using `SZ3 <https://github.com/szcompressor/SZ3>`_ and ZStd.
 * `HDF5-ZFP plugin <https://github.com/LLNL/H5Z-ZFP>`_ (v1.1.1)
   using `zfp <https://github.com/LLNL/zfp>`_ (v1.0.0).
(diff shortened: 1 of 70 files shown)


t20100:
The failing test is the one reading an already exisiting file, so that was compressed with a previous version of the filter.
The round-trip test pass.
So the new version of the SZ3 filter cannot read file generated with previous versions of the filter....

vasole:
> So the new version of the SZ3 filter cannot read file generated with previous versions of the filter....

I guess we'll have to check if the SZ3 developers are aware of it.

t20100:
This Pr is pending for a new release of SZ3 (see https://github.com/szcompressor/SZ3/issues/50#issuecomment-1915052871)

t20100:
Updated to use SZ3 v3.2.0 which checks that the data was compressed with a compatible version of the library.

Unfortunately, the `cd_values` format has changed from [a couple of arguments](https://github.com/silx-kit/hdf5plugin/blob/v4.4.0/doc/contribute.rst#sz3) to  the dump of an instance of a [Config](https://github.com/szcompressor/SZ3/blob/v3.2.0/include/SZ3/utils/Config.hpp) C++ class ([conversion code](https://github.com/szcompressor/SZ3/blob/b3dab4018425803a55d8073dc55dade7fa46b7b4/tools/H5Z-SZ3/src/H5Z_SZ3.cpp#L32-L71)).

This makes reading a HDF5 dataset compressed with a previous version of SZ3 abort the process because it cannot read the Config object. This defeats the check of the version by the library.
For instance, `test/test.py` fails with:
```
testSZ3 (__main__.TestHDF5PluginRead)
  Test reading and witing SZ3 compressed data ... terminate called after throwing an instance of 'std::length_error'
    what():  vector::_M_default_append
  start H5Z_filter_sz3
```

Also, when writing a new HDF5 dataset, the current passed arguments must be changed, so as it is the `hdf5plugin.test.TestHDF5PluginRW.testSZ3` test fails with:
```
start H5Z_sz3_set_local
terminate called after throwing an instance of 'std::length_error'
  what():  vector::_M_default_append
Aborted
```
It works when setting `SZ3.filter_options = ()`.
To make it work means re-implementing and the [dump of the Config object](https://github.com/szcompressor/SZ3/blob/b3dab4018425803a55d8073dc55dade7fa46b7b4/include/SZ3/utils/Config.hpp#L146-L178) to integers here in Python.
Since this mechanism was not developed to be use outside of C++, I'm -1 to duplicate it here.

Finally, it would be good to check that the Config read/write works when writing with one endianess and reading with another.

t20100:
See updates in https://github.com/szcompressor/SZ3/issues/50#issuecomment-2558747575

t20100:
Closing as outdate.

Attempt to update to SZ3 v3.3.0 is tracked in #351.

PR #290, opened 2024-01-24
SZ3 filter: Updated tests

t20100:
This PR updates SZ3 tests to relax the test to check that previously compressed data still respects the given constraints rather providing an exact match.

### DIFF
--- a/src/hdf5plugin/test.py
+++ b/src/hdf5plugin/test.py
@@ -438,6 +438,7 @@ class TestGetFilters(unittest.TestCase):
 class TestSZ(unittest.TestCase):
     """Specific tests for SZ compression"""
 
+    @unittest.skipUnless(should_test("sz"), "SZ filter not available")
     def testAbsoluteMode(self):
         """Test SZ's absolute mode is within required tolerance
 
--- a/test/test.py
+++ b/test/test.py
@@ -223,8 +223,6 @@ class TestHDF5PluginRead(unittest.TestCase):
                              "Values should not be identical")
             self.assertTrue(numpy.allclose(original, output_data, atol=value),
                              "Values should be within tolerance")
-            self.assertTrue(numpy.alltrue(compressed == output_data),
-                             "Compressed data should be identical")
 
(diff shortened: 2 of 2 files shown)


src/hdf5plugin/test.py
```diff
@@ -438,6 +438,7 @@ def testSelection(self):
 class TestSZ(unittest.TestCase):
     """Specific tests for SZ compression"""
 
+    @unittest.skipUnless(should_test("sz"), "SZ filter not available")
```
t20100, line 441: This was missing and it's convenient to have for testing without compiling all filters

test/test.py
```diff
@@ -223,8 +223,6 @@ def testSZ3(self):
                              "Values should not be identical")
             self.assertTrue(numpy.allclose(original, output_data, atol=value),
                              "Values should be within tolerance")
-            self.assertTrue(numpy.alltrue(compressed == output_data),
```
t20100, line 226: This is removed, other tests checks that both previously and newly compressed data respects the error bound from the original data.

test/test.py
```diff
@@ -269,7 +267,10 @@ def testSZ3(self):
             self.assertTrue(original.shape == compressed.shape, "Incorrect shape")
             self.assertFalse(numpy.alltrue(original == compressed),
                              "Values should not be identical")
-            self.assertTrue(numpy.allclose(compressed, compressed_back),
+            # Absolute error from L2 norm param from:
+            # https://github.com/szcompressor/SZ3/blob/v3.1.8/include/SZ3/utils/Statistic.hpp#L44
+            abs_error = numpy.sqrt(3.0/compressed.ndim) * value
```
t20100, line 272: Really unsure about this one, but it seems to be the way l2 norm argument is converted to abs error...
I also tried to check the L2 norm of the errors but it's way  above the given value.
Anyway, I'd take any better approach,

vasole: Indeed, I was expecting an L2 norm calculation somewhere...

PR #298, opened 2024-02-09
CI and Build: Updated setup.py and continuous integration configuration

t20100:
This PR aims at updating/simplifying the `setuptools` and CI configs and removing unnecessary code.

The main breaking change is the removal of `setup.py build` command line arguments that was deprecated in favor of environment variables.
Another change is the requirement of `setuptools>62.4.0` and `py-cpuinfo==9.0.0`, but that won't be visible when installing with `pip`.
I would include this in a major release.

Details of the changes:

- Requires `setuptools>62.4.0`: allow using `pyproject.toml` and removing fallbacks to `distutils`
- Move project metadata from `setup.py` to `pyproject.toml`
- Simplify CI test workflow and remove `ci/info_platform.py`
- Rename extras_require `[dev]` to `[doc]` and use it instead of `requirements.txt` to avoid duplicating the list of dependencies
- Remove local copy of `cpuinfo.py`: This is only needed when building with `python setup.py build` (as is `setuptools`). In this case, a log indicates to install `py-cpuinfo`.
- Remove supports for `setup.py build` command line arguments since this is a deprecated feature of `setuptools`.
- Rework some code in `setup.py`


t20100:
This PR also fixes support of `numpy` v2 in tests and in CI config.

Ready for review.

kif:
Indeed a tough one ...

pyproject.toml
```diff
@@ -1,3 +1,56 @@
 [build-system]
-requires = ["py-cpuinfo==8.0.0", "setuptools", "wheel>=0.34.0"]
+requires = ["py-cpuinfo==9.0.0", "setuptools>=62.4.0", "wheel>=0.34.0"]
 build-backend = "setuptools.build_meta"
+
+[project]
+name = "hdf5plugin"
+authors = [
+    {name = "ESRF - Data Analysis Unit", email = "silx@esrf.fr"}
```
kif, line 8: The name of the unit did change

kif:
Beside the 2 comments LGTM

kif (approved):
LGTM

pyproject.toml
vasole: I stopped putting Data Analysis Unit or Software Group and use just ESRF.

t20100: That sounds good, done

t20100:
Thanks for the review!

PR #299, opened 2024-03-14
Documentation: Fixed typo in `Bitshuffle` docstring

t20100:
Remove duplicated information in `Bitshuffle` docstring.

### DIFF
--- a/src/hdf5plugin/_filters.py
+++ b/src/hdf5plugin/_filters.py
@@ -115,8 +115,8 @@ class Bitshuffle(_FilterRefClass):
 
     :param int nelems:
         The number of elements per block.
-        It needs to be divisible by eight (default is 0, about 8kB per block)
-        Default: 0 (for about 8kB per block).
+        It needs to be divisible by eight.
+        Default: 0 (for about 8KB per block).
     :param str cname:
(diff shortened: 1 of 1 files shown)


src/hdf5plugin/_filters.py
kif, line 1: According to https://en.wikipedia.org/wiki/Kilobyte
1kB represents 1000 bytes, but KB is ambiguous.
One should write 8KiB for 8184B.

src/hdf5plugin/_filters.py
```diff
@@ -115,8 +115,8 @@ class Bitshuffle(_FilterRefClass):
 
     :param int nelems:
         The number of elements per block.
-        It needs to be divisible by eight (default is 0, about 8kB per block)
-        Default: 0 (for about 8kB per block).
+        It needs to be divisible by eight.
+        Default: 0 (for about 8KB per block).
```
t20100, line 119:
```suggestion
        Default: 0 (for about 8 kilobytes per block).
```

src/hdf5plugin/_filters.py
t20100: I went for `8 kilobytes` to lift all ambiguities

kif (approved):

PR #303, opened 2024-05-16
Added Sperr filter

### LINKED ISSUE #302: Add H5Z-SPERR to the collection?
Hello HDF5Plugin Developers,

[NCAR](https://en.wikipedia.org/wiki/National_Center_for_Atmospheric_Research) has an increased interest in `hdf5plugin` because it provides an easy approach to integrate lossy compression in Python-based workflows. However, `hdf5plugin` currently misses an important algorithm, namely [SPERR](https://github.com/NCAR/SPERR). I have recently developed a HDF5 plugin for `SPERR` ([H5Z-SPERR](https://github.com/NCAR/H5Z-SPERR/tree/v0.1.0)), and I wonder if it's possible to add `H5Z-SPERR` to `hdf5plugin`? I have very little knowledge in Python, but I can provide help in any other way w.r.t. the algorithm to make this integration happen.

Thanks,
Sam

t20100:
This PR adds the [H5Z-SPERR](https://github.com/NCAR/H5Z-SPERR) compression filters to `hdf5plugin`.

This is still WIP:

 - [x] no Windows support
 - [x] macos: Need at least macos10.13: I compiled with with `MACOSX_DEPLOYMENT_TARGET=10.15`
 - [x] Need more testing
 - [x] Seg fault for psnr and pwe modes on Debian10 and macos x86_64

closes #302


setup.py
```diff
@@ -458,10 +493,17 @@ def build_libraries(self, libraries):
...
-            build_info['cflags'] = [
+            cflags = [
                 f for f in cflags if f.startswith(prefix)
             ]
 
+            # Patch C++20 flag for older gcc/clang support
+            if '-std=c++20' in cflags:
+                if config.cpp20_compile_arg is None:
+                    raise RuntimeError("Cannot compile C++20 code with current compiler")
+                cflags = [f for f in cflags if f != '-std=c++20'] + [config.cpp20_compile_arg]
```
t20100, line 504: Allows to change `-std=c++20` to `-std=c++2a` for gcc and clang < 10

setup.py
```diff
@@ -683,6 +725,24 @@ def get_snappy_clib(field=None):
...
 
+def get_sperr_clib(field=None):
+    """sperr static lib build config"""
+    sperr_dir = "src/SPERR"
+
+    config = dict(
+        sources=glob(f"{sperr_dir}/src/*.cpp"),
+        include_dirs=[f"{sperr_dir}/include"],
+        macros=[
+            ("SPERR_VERSION_MAJOR", 0),
```
t20100, line 736: @shaomeng I had to set this macro which is used in the code.
Is there a better way?

setup.py
```diff
@@ -683,6 +725,24 @@ def get_snappy_clib(field=None):
     return config[field]
 
 
+def get_sperr_clib(field=None):
+    """sperr static lib build config"""
+    sperr_dir = "src/SPERR"
+
+    config = dict(
+        sources=glob(f"{sperr_dir}/src/*.cpp"),
```
t20100, line 733: @shaomeng, this builds all the src/ files, is it OK?

src/hdf5plugin/_filters.py
```diff
@@ -482,6 +486,112 @@ def __init__(self,
...
+
+        # encode mode in the top 4 bits
+        if mode == 1:
+            mask = 1 << (cls._INTEGER_BITS + cls._FRACTIONAL_BITS)
+        elif mode == 2:
+            mask = 1 << (cls._INTEGER_BITS + cls._FRACTIONAL_BITS)
+        else:  # mode == 3
+            mask = 1 << (cls._INTEGER_BITS + cls._FRACTIONAL_BITS)
+            mask |= 1 << (cls._INTEGER_BITS + cls._FRACTIONAL_BITS)
+        ret |= mask
```
t20100, line 590: This needs to be double-checked

src/hdf5plugin/_filters.py
```diff
@@ -482,6 +486,112 @@ def __init__(self,
...
+    filter_id = SPERR_ID
+
+    _FRACTIONAL_BITS = 16
+    _INTEGER_BITS = 12
+
+    def __init__(
+            self,
+            rate: float | None = None,
+            peak_signal_to_noise_ratio: float | None = None,
+            point_wise_error: float | None = None):
```
t20100, line 547: Instead of passing `mode` and `quality` as in the C filter, I choose to expose one argument for each mode as it is already done for `ZFP`, `SZ` and `SZ3` and tried to use similar names.

@shaomeng, from what I understood "point-wise error" tolerance is an absolute tolerance, am I right?
In this case I would rename it to `absolute` (as `SZ` and `SZ3`) or `accuracy` (as `Zfp`) (and it would have been best if we avoided this difference...)

t20100:
@shaomeng, first attempt at adding SPERR.

Still quite a few issues to tackle, in particular PSNR and PWE modes fails when running the tests (currently commented).
But cloning this branch and running `pip install .` you should be able to start testing it.

setup.py
shaomeng: `USE_VANILLA_CONFIG` is supposed to be used like this, so there's no problem.
`SPERR_VERSION_MAJOR` is populated by cmake from this line:
https://github.com/NCAR/SPERR/blob/ef135ee9d87fc9b64cdd213e93e03fedf151e7f5/CMakeLists.txt#L5
So I think if it's possible to extract that first digit (`0` right now) after the `VERSION` keyword, it'd be future proof. Otherwise, we probably need to manually update the `setup.py` once SPERR gets a version bump. I don't think it's gonna happen very often though.

setup.py
shaomeng: I don't quite understand this change, but I do want to point out that SPERR compiles with -std=c++17 too, on systems without C++20 support. (There's a small performance difference.)

setup.py
shaomeng: Yes, it should be building all the .cpp files.

src/hdf5plugin/_filters.py
shaomeng: Sorry that I changed added one more data field to be encoded in this function (you'll need to do a git pull to see the latest changes), so the function signature should look like:
```python
def __pack_options(cls, mode: int, quality: float, swap int) -> tuple[int]:
```
I have a few other comments:
- Line 586 should be: `mask = 1 << (cls._INTEGER_BITS + cls._FRACTIONAL_BITS + 1)`
- Line 589 should be: `mask |= 1 << (cls._INTEGER_BITS + cls._FRACTIONAL_BITS + 1)`
- Because of the added data field (`swap`) to be encoded, these lines are also needed:
```python
# encode the swap flag at bit 31
if swap != 0:
    ret |= 1 << (cls._INTEGER_BITS + cls._FRACTIONAL_BITS + 3)
```

src/hdf5plugin/_filters.py
shaomeng: It's correct that "point-wise error" is equivalent to "absolute error" so I'm happy to see it using that name too.

setup.py
t20100: OK, let's keep it manual for now

src/hdf5plugin/_filters.py
t20100: Thanks for the review! A second check was indeed needed!

src/hdf5plugin/_filters.py
t20100: Then I'll use `absolute` as keyword argument

src/hdf5plugin/_filters.py
t20100: This was making the tests fails (mode was always 1)

src/H5Z-SPERR/src/h5z-sperr.c
```diff
@@ -0,0 +1,347 @@
...
+  /*
+   * 	dcpl_id	  Dataset creation property list identifier
+   * 	type_id	  Datatype identifier
+   * 	space_id	Dataspace identifier
+   */
+
+  /* Get the user-specified compression mode and quality. */
+  size_t user_cd_nelem = 2, nchar = 16;
+  unsigned int user_cd_values[user_cd_nelem], flags;
+  char name[nchar];
```
t20100, line 162: Visual Studio is not happy with this code:
```
src/H5Z-SPERR/src/h5z-sperr.c(161): error C2057: expected constant expression
src/H5Z-SPERR/src/h5z-sperr.c(161): error C2466: cannot allocate an array of constant size 0
src/H5Z-SPERR/src/h5z-sperr.c(161): error C2133: 'user_cd_values': unknown size
src/H5Z-SPERR/src/h5z-sperr.c(162): error C2057: expected constant expression
src/H5Z-SPERR/src/h5z-sperr.c(162): error C2466: cannot allocate an array of constant size 0
src/H5Z-SPERR/src/h5z-sperr.c(162): error C2133: 'name': unknown size
src/H5Z-SPERR/src/h5z-sperr.c(204): warning C4244: '=': conversion from 'hsize_t' to 'unsigned int', possible loss of data
```

This should fix it:
```suggestion
  size_t user_cd_nelem = 2, nchar = 16;
  unsigned int user_cd_values[2], flags;
  char name[16];
```
... or maybe use of `const` :shrug: 

[full log](https://ci.appveyor.com/project/ESRF/hdf5plugin/builds/49866162/job/ro8kdw4am47l6uf5)

shaomeng: I have re-worked the code above, so if you do a git pull it should have my latest changes. If you think I should make a new minor release (`v0.1.3` for example) to include this change, I could do it too! 

I'll need to figure out how to run the python tests though :)

t20100: Thanks for the fix! I pulled the changes

A new minor release would be great, but maybe wait a bit too see if other such small changes are needed while testing hdf5plugin on different platforms.

Here is some documentation on how to run the tests once hdf5plugin is installed (`pip install .` or the old `python setup.py install`): https://hdf5plugin.readthedocs.io/en/stable/contribute.html#testing

To spare compilation time, you can disable the other filters by setting the `HDF5PLUGIN_STRIP` environment variable to `"bzip2,lz4,bshuf,blosc,blosc2,fcidecomp,zfp,sz,sz3,zstd"` when building hdf5plugin.

t20100:
@shaomeng, would you be able to provide us a HDF5 files with small datasets compressed with H5Z-SPERR (not compiled as part of `hdf5plugin`) along with the original data?

We have such files for most of the other filters in the [test/](https://github.com/silx-kit/hdf5plugin/tree/main/test) folder, so we check that `hdf5plugin` can read data compressed by the filters compiled outside of this project.

shaomeng:
> @shaomeng, would you be able to provide us a HDF5 files with small datasets compressed with H5Z-SPERR (not compiled as part of `hdf5plugin`) along with the original data?
> 
Sure! The attached file should be a 128x128 field, and it's actually a compressed version of [this](https://github.com/NCAR/H5Z-SPERR/blob/main/test_data/density_128x128.d64). Please let me know what more info I could provide! 
[sperr-example.zip](https://github.com/silx-kit/hdf5plugin/files/15422004/sperr-example.zip)

src/H5Z-SPERR/src/h5z-sperr.c
shaomeng: I followed the installation and test instructions, and it shows all tests are successful. I'm not sure what is exactly tested though...

t20100: This runs basic write/read tests for all available filters with a few different options and easy to compress data.
It is not meant to test the compression libraries itself but rather test that there is no obvious issue: it checks that some compression occurs and that the read-back data is the same (or close to) the original one.

https://github.com/silx-kit/hdf5plugin/blob/af093e455952df5f8aae5d6ced51cb1ee55805a2/src/hdf5plugin/test.py#L264-L275

shaomeng:
> > The attached file should be a 128x128 field, and it's actually a compressed version of [this](https://github.com/NCAR/H5Z-SPERR/blob/main/test_data/density_128x128.d64)
> 
> Great! Thanks a lot! I added a test with it.
> 
> Is it OK for you if this file is distributed in this repository under the MIT license?
> 

Yes, please feel free to included it in this repo. Thanks a lot!

t20100:
Added a test for macos min requirements to 10.13: The `-std=c++20` flag is accepted with 10.9 but some features required by SPERR are not available. 10.13 do not provide full support for C++20, but it is enough for what is required.

It now also uses macos 10.13 to build the release wheels, so macOS 10.13 (released september 2017) becomes the minimum supported version for the wheels.
It is still possible to build from source on older versions and SPERR will be skipped.

t20100:
This is ready for review and testing: From what I tested it works on all platform we typically run the tests on.

@shaomeng, thanks for the support! Let me know if/when you tag a release of h5z-sperr, we'll reference it in the documentation instead of a commit hash.
Otherwise, I usually check from new releases of the filters and compression libraries before making a new release of hdf5plugin.

shaomeng:
> This is ready for review and testing: From what I tested it works on all platform we typically run the tests on.
> 
> @shaomeng, thanks for the support! Let me know if/when you tag a release of h5z-sperr, we'll reference it in the documentation instead of a commit hash. Otherwise, I usually check from new releases of the filters and compression libraries before making a new release of hdf5plugin.

Thanks for the heads up! I have tagged a new release of `v0.1.3`: https://github.com/NCAR/H5Z-SPERR/releases/tag/v0.1.3
I'll run some test, hopefully in next week, and provide any feedback I have! 

Cheers!

t20100:
PR updated to embed H5Z-SPERR `v0.1.3`, thanks!

>Also, do you plan to release a new version of hdf5plugin soon? I'm considering tagging a new release of SPERR itself in the next 2 or 3 months. I definitely want to coordinate with you so the new hdfplugin release will have the latest SPERR.

There is no strict roadmap for `hdf5plugin`: On average there is a major/minor release every 6 months, but this depends on the availability of the maintainers. 

For now, I'm aiming at a major release this July or this autumn: [5.0.0](https://github.com/silx-kit/hdf5plugin/milestone/24).
If you tag a new version, do not hesitate to open an issue here and we'll update the embedded version.
I can also ping you when we plan to make the next release, and if you are about to release we can delay a bit to embedded the latest version.

Anyway, now that SPERR has been added to `hdf5plugin`, if there is no change in the HDF5 `cd_values` or new dependencies, it is easy to update the embedded version of the library and to make a release.

shaomeng (approved):

shaomeng:
> PR updated to embed H5Z-SPERR `v0.1.3`, thanks!
> 
> > Also, do you plan to release a new version of hdf5plugin soon? I'm considering tagging a new release of SPERR itself in the next 2 or 3 months. I definitely want to coordinate with you so the new hdfplugin release will have the latest SPERR.
> 
> There is no strict roadmap for `hdf5plugin`: On average there is a major/minor release every 6 months, but this depends on the availability of the maintainers.
> 
> For now, I'm aiming at a major release this July or this autumn: [5.0.0](https://github.com/silx-kit/hdf5plugin/milestone/24). If you tag a new version, do not hesitate to open an issue here and we'll update the embedded version. I can also ping you when we plan to make the next release, and if you are about to release we can delay a bit to embedded the latest version.
> 
> Anyway, now that SPERR has been added to `hdf5plugin`, if there is no change in the HDF5 `cd_values` or new dependencies, it is easy to update the embedded version of the library and to make a release.

Thanks for the info. Also, it took me a while to give it a review, but it looks really good and I've approved the PR. Thanks a lot for doing this!

kif (approved):
Impressive work

t20100:
Thanks for the reviews!

PR #305, opened 2024-05-28
CI: Updated release workflow to output wheel build logs

t20100:


t20100:
Using `macos-12` to build wheels since `macos-11` is about to be dropped (see https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources).
Use `macos-13` instead of `macos-latest` for CI to test on x86 instead of ARM.

t20100:
Closing as integrated in PR #303

PR #309, opened 2024-06-26
avoid throwing ValueError if user sets lz4=False

psteinb:
Just ran into this problem and would like to avoid other users to see a ValueError if they do:

```python
f = h5py.File('test.h5', 'w')
f.create_dataset(
    'bitshuffle_with_lz4',
    data=numpy.arange(100),
    **hdf5plugin.Bitshuffle(nelems=0, lz4=False, cname='zstd'))
f.close()
```


t20100 (approved):
Thanks for the fix!

PR #319, opened 2024-07-24
Prepare v5.0.0

t20100:
This PR prepares for the release of v5.0.

It bumps the release number and updates the changelog.
It also use `numpy<2` for tests during the release workflow since wheels of h5py v3.11 which supports numpy v2 are not available for `aarch64` and `ppc64le`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -104,7 +104,9 @@ jobs:
           CIBW_ARCHS: ${{ matrix.cibw_archs }}
 
           # Use silx wheelhouse for ppc64le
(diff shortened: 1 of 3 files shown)


t20100:
Preparing for a major release in a few days, unless there is any issue/extra feature or someone needs more time to have a look at it.


attn @shaomeng

shaomeng:
Thanks for the heads up. The only semi-significant change I made to SPERR since version 0.8.1 was to make use of a [C++20 feature](https://en.cppreference.com/w/cpp/numeric/countr_zero), and it results in a modest performance improvement. I think hdf5plugin was compiling SPERR using C++20 right? If so, I think I'll tag a release to be included!

t20100:
SPERR was updated and the build on Power9 has been fixed.

Unless there is other fix/updates needed, I'll make the release tomorrow.

payno (approved):

Issue #328, opened 2025-03-05
[5.0.0][debian] IndexError: list index out of range


picca:
Hello, I try to release the 5.0.0 vesion of hdf5plugin but I get this error message during the build

```
dh clean --buildsystem=pybuild
   dh_auto_clean -O--buildsystem=pybuild
I: pybuild base:311: python3.13 setup.py clean 
Traceback (most recent call last):
  File "/<<PKGBUILDDIR>>/setup.py", line 1271, in <module>
    get_zlib_clib(),
    ~~~~~~~~~~~~~^^
  File "/<<PKGBUILDDIR>>/setup.py", line 818, in get_zlib_clib
    zlib_dir = glob('src/c-blosc/internal-complibs/zlib*')[0]
               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^
IndexError: list index out of range
E: pybuild pybuild:389: clean: plugin distutils failed with: exit code=1: python3.13 setup.py clean 
dh_auto_clean: error: pybuild --clean -i python{version} -p 3.13 returned exit code 13
make: *** [debian/rules:8: clean] Error 25
```

I requested

```
export HDF5PLUGIN_STRIP=all
```

but it still try to find some library sources.

t20100:
The static libraries configs are retrieved before the "STRIP":

https://github.com/silx-kit/hdf5plugin/blob/a1a676fbee76f3f1cf60b7895ace5ed9a0f89675/setup.py#L1265-L1290

We should be able to define the static libs configs after the "STRIP". This would solve this issue.

t20100:
https://github.com/silx-kit/hdf5plugin/pull/331/commits/b0ee98105e706471ce54136ee0543d948fbab0d2 should fix your issue.

t20100:
@picca I'm updating `hdf5plugin` and I'll make a new release shortly. This should not add features for Debian since updates concerns embedded filters but it will fix this issue.

PR #333, opened 2025-03-14
Build: Added `HDF5PLUGIN_SYSTEM_LIBRARIES` env. var. to build using system libraries

### LINKED ISSUE #324: Add a way to build with system libraries
First raised in https://github.com/conda-forge/hdf5plugin-feedstock/issues/55

The idea is to allow using system libraries for the compression libs instead of the embedded ones.
The [request comes from conda-forge](https://github.com/conda-forge/hdf5plugin-feedstock/issues/55), but OpenSUSE is also using this approach through [a patch](https://src.opensuse.org/pool/python-hdf5plugin/src/branch/factory/hdf5plugin-system-libs.patch).

I propose to use a `HDF5PLUGIN_SYSTEM_LIBS` environment variable that sets the list of system libs to use (the value can also be `all`) as it is done for the `HDF5PLUGIN_STRIP` env. var.

### LINKED ISSUE #332: hdf5plugin build against dynamic system libraries instead of embedded static ones
`gcc` favors dynamic libs over static ones when both are available.

For instance, when building from source, if `lz4` dynamic library is available on the system it will be used instead of the statically compiled one since Python adds `-llz4` to the link command and `gcc` links with the dynamic lib from the system

t20100:
This PR reworks the management of C libraries in `setup.py` to:
- Allow to use system libraries instead of embedded sources, closes #324
- Run embedded plugins/libraries configuration code only if they actually are needed, related to #328
- Fix link with dynamic system libs instead of static ones, closes #332

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
-# Copyright (c) 2016-2024 European Synchrotron Radiation Facility
+# Copyright (c) 2016-2025 European Synchrotron Radiation Facility
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
@@ -734,7 +734,8 @@ def get_lz4_clib(field=None):
     cflags = ['-O3', '-ffast-math', '-std=gnu99']
     cflags += ['/Ox', '/fp:fast']
 
-    lz4_dir = glob('src/c-blosc2/internal-complibs/lz4*')[0]
+    folders = glob('src/c-blosc2/internal-complibs/lz4*')
+    lz4_dir = folders[0] if folders else "src/no-folder"
 
     config = dict(
         sources=glob(f'{lz4_dir}/*.c'),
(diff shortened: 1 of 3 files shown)


t20100:
Tested with conda: no longer link with available system libs if not requested + use them if requested through the `HDF5PLUGIN_SYSTEM_LIBRARIES` env. var. (the corresponding pkg-config .pc file must be available).

vasole:
I have only found things like this:

https://stackoverflow.com/questions/42585210/extending-setuptools-extension-to-use-cmake-in-setup-py

setup.py
```diff
@@ -666,7 +671,40 @@ def prefix(directory, files):
...
+            "-O2",
+            "-g",
+            "-D_FILE_OFFSET_BITS=64"
+        ],
+    )
+
+    if field is None:
+        return config
+    if field in ('extra_link_args', 'libraries'):
+        return []
```
payno, line 703: 'side question' not sure I get this part. Why is this returning a list ? I would expect an empty dict.

payno (approved):

setup.py
t20100: If `field` is `None`, it returns a `dict` which contains the build config for the static lib which is passed to `BuildCLib` setuptools command.
If `field` is provided, it returns the value of the field. It is called with way from the hdf5 plugin code using the compression lib to get what needs to be passed for linking with the compression lib to the setuptools `Extension`. `Extension` expects `list`s for 'extra_link_args' and 'libraries'.

It cannot be all stored in a `dict` since 'extra_link_args' and 'libraries' are not expected by `BuildCLib`.
It would probably be best to use a class/dataclass/namedtuple here, or a function to extract what's relevant to `BuildCLib`. Here I've minimised the rewrite, but I agree it makes things complex

t20100:
Let's merge PR #335 first

PR #335, opened 2025-03-17
Updated compression filer `h5z-sperr` v0.2.3

t20100:
This MR updates h5z-sperr to version ~~[0.2.1](https://github.com/NCAR/H5Z-SPERR/releases/tag/v0.2.1)~~ [0.2.3](https://github.com/NCAR/H5Z-SPERR/releases/tag/v0.2.3).
A new `cd_values` is added to pass ["missing_value_mode"](https://github.com/NCAR/H5Z-SPERR/tree/v0.2.1?tab=readme-ov-file#handling-of-missing-values)

### DIFF
--- a/doc/information.rst
+++ b/doc/information.rst
@@ -77,7 +77,7 @@ HDF5 compression filters and compression libraries sources were obtained from:
 * `SZ plugin <https://github.com/szcompressor/SZ>`_
   (commit `f466775 <https://github.com/szcompressor/SZ/tree/f4667759ead6a902110e80ff838ccdfddbc8dcd7>`_)
   using `SZ <https://github.com/szcompressor/SZ>`_, ZLib and ZStd.
-* `H5Z-SPERR plugin <https://github.com/NCAR/H5Z-SPERR>`_ (v0.1.3) using `SPERR <https://github.com/NCAR/SPERR>`_ (v0.8.2).
+* `H5Z-SPERR plugin <https://github.com/NCAR/H5Z-SPERR>`_ (v0.2.3) using `SPERR <https://github.com/NCAR/SPERR>`_ (v0.8.2).
 * `SZ3 plugin <https://github.com/szcompressor/SZ3>`_
   (commit `4bbe9df7e4bcb <https://github.com/szcompressor/SZ3/commit/4bbe9df7e4bcb6ae6339fcb3033100da07fe7434>`_)
   using `SZ3 <https://github.com/szcompressor/SZ3>`_ and ZStd.
--- a/setup.py
+++ b/setup.py
@@ -773,12 +773,13 @@ def get_snappy_clib(field=None):
 
 
 def get_sperr_clib(field=None):
-    """sperr static lib build config"""
+    """sperr static lib and sperr filter C++20 source file build config"""
     sperr_dir = "src/SPERR"
+    h5z_sperr_dir = "src/H5Z-SPERR"
 
     config = dict(
-        sources=glob(f"{sperr_dir}/src/*.cpp"),
-        include_dirs=[f"{sperr_dir}/include"],
+        sources=glob(f"{sperr_dir}/src/*.cpp") + [f"{h5z_sperr_dir}/src/h5zsperr_helper.cpp"],
+        include_dirs=BuildConfig.get_hdf5_include_dirs() + [f"{sperr_dir}/include", f"{h5z_sperr_dir}/include"],
         macros=[
             ("SPERR_VERSION_MAJOR", 0),  # Check project(SPERR VERSION ... in src/SPERR/CMakeLists.txt
             ("USE_VANILLA_CONFIG", 1),
@@ -1223,7 +1224,14 @@ def get_sperr_plugin():
 
     return HDF5PluginExtension(
         "hdf5plugin.plugins.libh5sperr",
-        sources=[f"{h5z_sperr_dir}/src/h5z-sperr.c"],
+        sources=prefix(
+            f"{h5z_sperr_dir}/src",
+            [
+                "h5z-sperr.c",
+                "icecream.c",
+                "compactor.c",
+            ]
+        ),
         include_dirs=get_sperr_clib("include_dirs") + [f"{h5z_sperr_dir}/include"],
         extra_link_args=['-lstdc++'],
         define_macros=get_sperr_clib("macros"),
--- a/setup.py
+++ b/setup.py
@@ -786,6 +786,11 @@ def get_sperr_clib(field=None):
         ],
         cflags=["-std=c++20", "/std:c++20"],
     )
+
+    if sys.platform == 'win32':
+        # h5zsperr_helper.cpp is part of the plugin
+        config["macros"].append(('H5_BUILT_AS_DYNAMIC_LIB', None))
+
     if field is None:
         return 'sperr', config
     return config[field]
--- a/src/H5Z-SPERR/.gitignore
+++ b/src/H5Z-SPERR/.gitignore
@@ -1,3 +1,6 @@
+# vim temp files
+*.sw?
+
 # Prerequisites
 *.d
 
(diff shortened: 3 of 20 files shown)


t20100:
@shaomeng, I found the follwing issue when trying to update h5z-sperr:

This test:
https://github.com/silx-kit/hdf5plugin/blob/4eaa8f26cbfcaad9f7f0792348d729cd9f1345a4/test/test.py#L148-L163

Leads to:
```
  testSperr (__main__.TestHDF5PluginRead)
  Test reading Sperr compressed data ... sh: line 1:  1278 Segmentation fault      (core dumped) python /project/test/test.py
```

Reading [this file](https://github.com/silx-kit/hdf5plugin/blob/4eaa8f26cbfcaad9f7f0792348d729cd9f1345a4/test/sperr.h5) works with h5z-sperr v0.1.3 but seg fault when updating to h5z-sperr v0.2.1.

t20100:
There is also a build issue in the hdf5plugin project on Windows:
```
    INFO:root:"C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.29.30133\bin\HostX86\x64\link.exe" /nologo /INCREMENTAL:NO /LTCG /DLL /MANIFEST:EMBED,ID=2 /MANIFESTUAC:NO /LIBPATH:src/hdf5/lib /LIBPATH:C:\Users\runneradmin\AppData\Local\Temp\cibw-run-oegt6uv8\cp38-win_amd64\build\venv\libs /LIBPATH:C:\Users\runneradmin\AppData\Local\pypa\cibuildwheel\Cache\nuget-cpython\python.3.8.10\tools\libs /LIBPATH:C:\Users\runneradmin\AppData\Local\pypa\cibuildwheel\Cache\nuget-cpython\python.3.8.10\tools /LIBPATH:C:\Users\runneradmin\AppData\Local\Temp\cibw-run-oegt6uv8\cp38-win_amd64\build\venv\PCbuild\amd64 /LIBPATH:build\temp.win-amd64-cpython-38 "/LIBPATH:C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.29.30133\ATLMFC\lib\x64" "/LIBPATH:C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.29.30133\lib\x64" "/LIBPATH:C:\Program Files (x86)\Windows Kits\NETFXSDK\4.8\lib\um\x64" "/LIBPATH:C:\Program Files (x86)\Windows Kits\10\lib\10.0.22621.0\ucrt\x64" "/LIBPATH:C:\Program Files (x86)\Windows Kits\10\lib\10.0.22621.0\um\x64" hdf5.lib charls.lib lz4.lib snappy.lib sperr.lib zfp.lib zlib.lib zstd.lib /EXPORT:H5PLget_plugin_info build\temp.win-amd64-cpython-38\Release\src/H5Z-SPERR/src/compactor.obj build\temp.win-amd64-cpython-38\Release\src/H5Z-SPERR/src/h5z-sperr.obj build\temp.win-amd64-cpython-38\Release\src/H5Z-SPERR/src/icecream.obj /OUT:build\lib.win-amd64-cpython-38\hdf5plugin\plugins\libh5sperr.dll /IMPLIB:build\temp.win-amd64-cpython-38\Release\src/H5Z-SPERR/src\libh5sperr.lib
       Creating library build\temp.win-amd64-cpython-38\Release\src/H5Z-SPERR/src\libh5sperr.lib and object build\temp.win-amd64-cpython-38\Release\src/H5Z-SPERR/src\libh5sperr.exp
    sperr.lib(h5zsperr_helper.obj) : error LNK2001: unresolved external symbol H5E_BADVALUE_g
    sperr.lib(h5zsperr_helper.obj) : error LNK2001: unresolved external symbol H5E_ERR_CLS_g
    sperr.lib(h5zsperr_helper.obj) : error LNK2001: unresolved external symbol H5E_PLINE_g
    build\lib.win-amd64-cpython-38\hdf5plugin\plugins\libh5sperr.dll : fatal error LNK1120: 3 unresolved externals
```

shaomeng:
Hi @t20100 , thank you so much for updating hdf5plugin!

Regarding the linking error, I think it's just about a header, and the `0.2.2` release (https://github.com/NCAR/H5Z-SPERR/releases/tag/v0.2.2) should have fixed it. Could you give it a try? 

Regarding the failed test, it's because that I broke backward compatibility... Does it mean that if I supply a few new test files, then we can have the tests pass?

t20100:
> Regarding the linking error, I think it's just about a header, and the 0.2.2 release (https://github.com/NCAR/H5Z-SPERR/releases/tag/v0.2.2) should have fixed it. Could you give it a try?

It does not fix the issue. But the problem is in hdf5plugin that needs to provide the missing symbols to the lib instead of libhdf5.

> Regarding the failed test, it's because that I broke backward compatibility... Does it mean that if I supply a few new test files, then we can have the tests pass?

The main problem is users won't be able to read older data and have their application to crash instead without an error message because they updated this package...
I was planning to make a minor release, so a breaking change won't go into it.
Also it would be great to detect older/unsupported compressed data and show an explicit error message in this case if it's not possible to keep the backward compatibility.

shaomeng:
> It does not fix the issue. But the problem is in hdf5plugin that needs to provide the missing symbols to the lib instead of libhdf5.

Thanks for clarifying. Does it mean that I don't need to do anything to have it fixed? 

> 
> The main problem is users won't be able to read older data and have their application to crash instead without an error message because they updated this package... I was planning to make a minor release, so a breaking change won't go into it. Also it would be great to detect older/unsupported compressed data and show an explicit error message in this case if it's not possible to keep the backward compatibility.

This makes sense. I've changed the code to check "compatibility versions" and raise a proper HDF5 error when they don't match, instead of going into segfaults. Would you mind giving the `head` of the main branch a try? I can tag a new release if it behaves as expected.

shaomeng:
Hi @t20100 , I actually just implemented the ability to read files produced by earlier versions of `H5Z-SPERR`; it's not too bad ;) I've also tagged a new release, version `0.2.3`, that includes this change. Do you mind giving it a test, and possibly make it into your next release?

t20100:
@shaomeng v0.2.3 passes the tests fine with the older sperr compressed hdf5 file, thanks!

I still have to fix the build issue on Windows: the problem is the build a plugin with a mixture of C and C++20 files with `setuptools`. I'll find a way

shaomeng:
> @shaomeng v0.2.3 passes the tests fine with the older sperr compressed hdf5 file, thanks!

Great! Do you think I need to supply a hdf5 that's compressed using the new version? 
> 
> I still have to fix the build issue on Windows: the problem is the build a plugin with a mixture of C and C++20 files with `setuptools`. I'll find a way

If it helps, the C++ code only requires C++14, not C++20. I'm not sure if it makes things easier for you. Sorry :)

t20100:
> Do you think I need to supply a hdf5 that's compressed using the new version?

If you a small have one, that'd be great, we can add it to the test files: https://github.com/silx-kit/hdf5plugin/tree/main/test

Windows compilation fixed by adding a macro to the C++ build part to declare hdf5 symbols as external while building sperr as a static lib.

Ready for review.

@shaomeng, would you mind having a look at the `missing_value_mode` argument support : https://github.com/silx-kit/hdf5plugin/pull/335/commits/f294eb7ca16ce1f51449c67ab248ef419d8dab7b

payno (approved):

shaomeng:
> 
> If you a small have one, that'd be great, we can add it to the test files: https://github.com/silx-kit/hdf5plugin/tree/main/test
> 
> @shaomeng, would you mind having a look at the `missing_value_mode` argument support : [f294eb7](https://github.com/silx-kit/hdf5plugin/commit/f294eb7ca16ce1f51449c67ab248ef419d8dab7b)

I've created this test file:
[sperr_missing_val.zip](https://github.com/user-attachments/files/19458754/sperr_missing_val.zip)
The thing to verify upon decompression is that locations at the indexes of `(1, 10, 100)` are `NaN`s, while other locations have normal 64-bit floating point values. 

Also, the missing value mode support looks good to me; thanks for making it happen!

PR #344, opened 2025-08-25
Doc: Updated license syntax in `pyproject.toml`

t20100:
This PR updates the license info to avoid a warning during build.

closes #343

PR #345, opened 2025-08-28
Updated minimum required version of Python

t20100:
This PR updates the minimum required version of Python to 3.9.
This is needed for PR #344 which is needed so that it keeps working with `setuptools` released after February 2026...

The CI configs are updated to add python3.13 and remove 3.8.
This PR also updates those CI configs

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,7 +9,7 @@ authors = [
 ]
 description = "HDF5 Plugins for Windows, MacOS, and Linux"
 readme = "README.rst"
-requires-python = ">=3.8"
+requires-python = ">=3.9"
 license = {file = "LICENSE"}
 classifiers = [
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -44,8 +44,9 @@ doc = [
     "sphinx_rtd_theme",
 ]
 test = [
-    "blosc2>=2.5.1;python_version>='3.9'",
-    "blosc2-grok>=0.2.2;python_version>='3.9'",
+    "numpy<2 ; python_version == '3.10'",
```
t20100, line 47: Use `numpy` v1 for Python3.9: `blosc2 wheel does not support numpy 2 but does not advertise it

payno:
Fine with it. I guess the changelog will be updated in another PR.

PR #352, opened 2025-09-16
Marked SZ filter as deprecated

t20100:
This PR updates SZ filter links since the repo was renamed and mark the SZ as deprecated as stated in https://github.com/szcompressor/SZ


src/hdf5plugin/_filters.py
```diff
@@ -654,7 +654,9 @@ class SZ(h5py.filters.FilterRefBase):
               data=numpy.random.random(100),
               compression=hdf5plugin.SZ(pointwise_relative=0.01))
 
-    For more details about the compressor, see `SZ compressor <https://github.com/szcompressor/SZ>`_.
+    For more details about the compressor, see `SZ2 compressor <https://github.com/szcompressor/SZ2>`_.
+
+    .. warning:: The SZ2 compressor is deprecated, see `SZ <https://github.com/szcompressor/SZ>`_
```
payno, line 659: I don't know how safe it is to provide a specific section. This might break. But at worst it should fall back on the page.
```suggestion
    .. warning:: The SZ2 compressor is deprecated, see `SZ <https://github.com/szcompressor/SZ?tab=readme-ov-file#deprecatedlegacy-version-sz2>`_
```

Issue #351, opened 2025-09-16
Update compression libraries


t20100:
Compression libs and filters updates to consider:

- [x] https://github.com/google/snappy/releases/tag/1.2.2
- [x] https://github.com/NCAR/SPERR/releases/tag/v0.8.3 (Added AVX2 support)
- [x] https://github.com/Blosc/c-blosc2/releases/tag/v2.21.2
- [ ] https://github.com/szcompressor/SZ3/releases/tag/v3.3.0

vasole:
If upstream confirms it is not a bug and it is accepted behavior at their side, we should upgrade.

PR #359, opened 2025-09-24
Updated SZ3 to v3.3.0

t20100:
This PR aims at updating SZ3 to the current latest version (3.3.0).

As it is, it seg fault when reading data written with the previous version of the library:

`python test/test.py` terminate with:

```
testSZ3 (__main__.TestHDF5PluginRead.testSZ3)
libc++abi: terminating with uncaught exception of type std::invalid_argument: magic number mismatch, the input data is not compressed by SZ3
```

This is due to an incompatible change in the `cd_values` stored in the hdf5 file.

With v3.3.0, the `cd_values` are a dump of "SZ3::Config", saved by [Config::save](https://github.com/szcompressor/SZ3/blob/5d2fd2638bc3512c0b4aed594d6e8a130d0d3ed2/include/SZ3/utils/Config.hpp#L221).
See those links for save to/load from hdf5 cd_values:
- https://github.com/szcompressor/SZ3/blob/5d2fd2638bc3512c0b4aed594d6e8a130d0d3ed2/tools/H5Z-SZ3/src/H5Z_SZ3.cpp#L36
- https://github.com/szcompressor/SZ3/blob/5d2fd2638bc3512c0b4aed594d6e8a130d0d3ed2/tools/H5Z-SZ3/src/H5Z_SZ3.cpp#L69

Related to #351


t20100:
The format of `cd_values` has changed, this implies either:

- to call C++ code to instantiate&set-up `SZ3::Config` (example of conf init: https://github.com/szcompressor/SZ3/blob/635ddc356b765967f1074a12574b1895d70fe9b9/tools/sz3c/src/sz3c.cpp#L14-L41) and call [Config::save](https://github.com/szcompressor/SZ3/blob/5d2fd2638bc3512c0b4aed594d6e8a130d0d3ed2/include/SZ3/utils/Config.hpp#L221). I'm really :-1: on this because the package would no longer be "pure" Python

- reimplement this `Config::save` in Python, but provided the amount of parameters, this is quite error-prone

PR #360, opened 2025-10-07
Simplify typing

t20100:
Updated code with `pyupgrade --py39-plus`


t20100:
Not on silx yet, but on other projects yes!

PR #379, opened 2026-03-16
Fix: Low performance for lz4 decompression

### LINKED ISSUE #326: Performance regression for lz4 decompression after version version 4.0.1
I noticed that the LZ4 decompression via hdf5plugin 4.1.0 and later is much slower than with hdfplugin 4.0.1, while the compression speed is very similar. I would have expected a substantial speedup when using version 5.0 with update to lz4 1.10 with the new multithreaded decompression (or is that not used at all)., but the decompression speed for 4.1.x and 5.0 seems to be the same:

```python
import time
import hdf5plugin
import h5py
import numpy as np
from io import BytesIO

test_data = np.ones((1024, 1024, 1024), np.uint8)

raw_buffer = BytesIO()

with h5py.File(raw_buffer, 'w') as f:
    compression_start_time = time.perf_counter()
    f.create_dataset('data', data=test_data, compression=hdf5plugin.LZ4())
    compression_time = time.perf_counter() - compression_start_time

with h5py.File(raw_buffer, 'r') as f:
    decompression_start_time = time.perf_counter()
    data = f['data'][:]
    decompression_time = time.perf_counter() - decompression_start_time

print(f"hdf5plugin {hdf5plugin.version}: "
      f"lz4 compression time {compression_time:.3f}s, "
      f"lz4 decompression_time: {decompression_time:.3f}s")
```

gives the following results for different hdf5plugin version with h5py 3.12.1 on Python 3.11.9 on Windows 10 (AMD Ryzen 7 5900X):

> hdf5plugin 4.0.1: lz4 compression time 0.219s, lz4 decompression_time: 0.283s
> hdf5plugin 4.1.0: lz4 compression time 0.226s, lz4 decompression_time: 1.630s
> hdf5plugin 5.0.0: lz4 compression time 0.221s, lz4 decompression_time: 1.610s

I have seen similar results on Python 3.8 and 3.11 on Debian 12 with different h5py versions.

DominiqueCEOS:
With the LZ4 version included in the pypi released with the lz4 decompression with hdf5plugin 4.1.0 - 6.0.0 via decompression via the lz4_decompress_fast method got substantially slower then in hdf5plugin < 4.1.0.
lz4 decompress_fast is also deprecated in the LZ4 1.9.0 release since it is unsafe and shall be replaced by lz4_decompress_safe.

Note: this drops support for older LZ4 versions or at least might slow down the decompression when using lz4 libraries < 1.9.0, which is irrelevant to the currently builds wheels that are released to pypi.

Resolves #326

### DIFF
--- a/lib/LZ4/H5Zlz4.c
+++ b/lib/LZ4/H5Zlz4.c
@@ -99,14 +99,15 @@ static size_t H5Z_filter_lz4(unsigned int flags, size_t cd_nelmts,
             }
             else /* do the decompression */
             {
-#if LZ4_VERSION_NUMBER > 10300
-                int compressedBytes = LZ4_decompress_fast(rpos, roBuf, blockSize);
-#else
-                int compressedBytes = LZ4_uncompress(rpos, roBuf, blockSize);
-#endif
-                if(compressedBytes != compressedBlockSize)
+                int decompressedBytes = LZ4_decompress_safe(
+                        rpos,
+                        roBuf,
+                        compressedBlockSize,
+                        blockSize);
+                if (decompressedBytes < 0 || decompressedBytes != blockSize)
                 {
-                    printf("decompressed size not the same: %d, != %d\n", compressedBytes, compressedBlockSize);
+                    printf("decompression failed or wrong size: %d != %u\n",
+                           decompressedBytes, blockSize);
                     goto error;
                 }
             }


t20100:
Hi,

Thanks for opening a Pull Request!

However, in this project we are embedding hdf5 compression filters but not maintaining them.
So we aim at not patching the compression filters unless there is at least a pull request on the upstream project (and better if merged).

You can propose your change to the upstream project https://github.com/nexusformat/HDF5-External-Filter-Plugins but this project looks stalled.

The HDFGroup maintains a [hdf5_plugins](https://github.com/HDFGroup/hdf5_plugins) project where the LZ4 filter already uses `LZ4_decompress_safe`, see https://github.com/HDFGroup/hdf5_plugins/blob/edf3857d97deb635117aa1cb2d2a92c31ac82f4d/LZ4/src/H5Zlz4.c#L147-L151

In #326 I propose to switch to this upstream project instead which should improve the performances.

I'm closing this PR but keeps the related issue opened.

PR #380, opened 2026-03-28
HDF5-2 compatibility - always link to hdf5 library

hmaarrfk:
In building for HDF5-2 compatibility at conda-forge, i found that globals like `H5_libinit_g` now need the library to be linked.

https://github.com/conda-forge/hdf5plugin-feedstock/pull/72

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -682,7 +682,7 @@ class HDF5PluginExtension(Extension):
             self.export_symbols.append("init_filter")
         else:
             self.define_macros.append(("H5_BUILT_AS_DYNAMIC_LIB", None))
-            self.libraries.append("hdf5")
+        self.libraries.append("hdf5")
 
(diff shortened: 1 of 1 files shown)


t20100:
Thanks for the pushing the fix upstream!

`H5_BUILT_AS_DYNAMIC_LIB` macro may also need to be set, to be checked.

So far on Windows in was linking missing symbols to already loaded libs.

t20100:
Thanks for proposition this fix!

Digging into this issue, this PR fixes works well for conda-forge but is not appropriate when building a wheel of `hdf5plugin` since the hdf5 library is not known at compile time.

I'm closing this PR in favor of PR #391, which is expected to work both for building in conda-forge and building the wheels.

hmaarrfk:
thx for looking into this

PR #381, opened 2026-04-24
Add `__all__` to `__init__.py` for type checker re-export compatibility

riccardoratta:
## Summary

Pyright/Pylance (and other PEP 484-compliant type checkers) do not treat `from .module import Name` in a package `__init__.py` as a public re-export. As a result, users of `hdf5plugin` with strict type checking enabled see errors such as:

```
"Blosc2" is not exported from module "hdf5plugin"
```

even though the import works correctly at runtime.

## Root Cause

Pyright follows the rule defined in the [typing spec](https://typing.readthedocs.io/en/latest/spec/distributing.html#library-interface): a name imported in a package's `__init__.py` is only considered a public re-export if it is:
- re-imported using the `from x import y as y` form, or
- listed in `__all__`.

The existing `# noqa` comments suppress linter warnings (e.g. flake8 F401) but have no effect on type checker re-export analysis.

## Change

Update all public re-exports in `__init__.py` to use the explicit `from ._filters import Blosc2 as Blosc2` form, which satisfies both type checkers and linters without requiring an `__all__` declaration.

**It has no runtime behaviour changes**.


src/hdf5plugin/__init__.py
```diff
@@ -61,6 +61,40 @@
...
+    "FCIDECOMP_ID",
+    "LZ4_ID",
+    "SPERR_ID",
+    "SZ_ID",
+    "SZ3_ID",
+    "ZFP_ID",
+    "ZSTD_ID",
+    # Utilities
+    "FILTERS",
+    "PLUGIN_PATH",
```
t20100, line 79: `from_filter_options` is missing:

```suggestion
    "PLUGIN_PATH",
    "from_filter_options",
```

src/hdf5plugin/__init__.py
```diff
@@ -61,6 +61,40 @@
...
+    "BLOSC2_ID",
+    "BSHUF_ID",
+    "BZIP2_ID",
+    "FCIDECOMP_ID",
+    "LZ4_ID",
+    "SPERR_ID",
+    "SZ_ID",
+    "SZ3_ID",
+    "ZFP_ID",
+    "ZSTD_ID",
```
t20100: I propose to remove the `*_ID` constants from `__all__`: The ids are accessible from the filter classes, e.g., `Blosc2.filter_id`, and I would consider deprecating the `*_ID` constants (not in this PR).

```suggestion
```

t20100 (commented):
Thanks for the contribution!

riccardoratta:
Done, I've added `from_filter_options` and removed the IDs.

PR #383, opened 2026-05-19
Updated SPERR compression library (v0.8.5) and blosc2 hdf5 filter.

t20100:
This PR updates:
- SPERR to the latest bug fix release
- hdf5-blosc2 compression filter to the tip of the main branch to include a fix required for building with c-blosc2 v3.

The only manually changed file is `doc/information.rst`, the rest is the result of `git subtree pull ...` commands.

related to #382 and https://github.com/conda-forge/hdf5plugin-feedstock/pull/74

PR #384, opened 2026-05-19
Bitshuffle: Fixed `clevel` argument supported range

t20100:
This PR corrects the docstring and adds a check of the passed `clevel` value (see https://github.com/silx-kit/hdf5plugin/issues/375#issuecomment-4488083987 for description of the issue with negative values).

It also adds a test with different clevel values.

closes #375


t20100:
pinning blosc2-grok and blosc2 for tests to pass as in PR #383

payno (approved):

PR #394, opened 2026-06-23
Updated LZ4 plugin

### LINKED ISSUE #326: Performance regression for lz4 decompression after version version 4.0.1

t20100:
This PR changes upstream project for the LZ4 from the initial repository (https://github.com/nexusformat/HDF5-External-Filter-Plugins) to the version hosted by the hdfgroup organisation (https://github.com/HDFGroup/hdf5_plugins).

The reason is that the former project is stalled while the later is actively maintained.
As raised in #326 this has consequences on performances.

Running the sample code from https://github.com/silx-kit/hdf5plugin/issues/326#issue-2622368995 shows performance improvement:
```
hdf5plugin 6.0.0: lz4 compression time 0.270s, lz4 decompression_time: 1.577s
hdf5plugin 7.0.0: lz4 compression time 0.252s, lz4 decompression_time: 0.365s
```

closes #326


t20100:
Only copying a small part of `hdf5_plugins` repository rather than using `git subtree` since a small part of it is used for now. It was already done this way for the bzip2 filter (rather than integrating the whole pytables repo here).

In the future, we should consider relying on the hdf group's hdf5_plugins repository.

payno (approved):

PR #165, opened 2022-03-04
Allow for negative clevels in Zstd

mkitti:
This an extension of #164 to modify the Zstd plugin to allow for negative compression levels. Upstream pull request is linked below. There have been no changes to upstream in five years and no activity from the author for almost three years.

Upstream: https://github.com/aparamon/HDF5Plugin-Zstandard/pull/8

PR #395, opened 2026-06-23
Updated Zstd plugin; Added support for negative compression levels to Zstd

t20100:
! Merge PR #394 first !

This PR changes upstream project for the Zstd filter from the initial repository (ttps://github.com/aparamon/HDF5Plugin-Zstandard) to the version hosted by the hdfgroup organisation (https://github.com/HDFGroup/hdf5_plugins).

The reason is that the former project is stalled while the later is actively maintained.
For instance, support for negative compression levels (favor speed over compression ratio) as initially proposed in #165  is implemented in the hdfgroup's version of the filter.

This PR also adds support for negative compression levels in `hdf5plugin.Zstd` class.


src/hdf5plugin/_filters.py
```diff
@@ -1270,25 +1270,35 @@ class Zstd(FilterBase):
...
+    _ZSTD_MAX_CLEVEL = 22
+
     def __init__(self, clevel: int = 3):
-        if not 1 <= clevel <= 22:
-            raise ValueError("clevel must be in the range [1, 22]")
+        if not self._ZSTD_MIN_CLEVEL <= clevel <= self._ZSTD_MAX_CLEVEL:
+            raise ValueError(
+                f"clevel must be in the range [{self._ZSTD_MIN_CLEVEL}, {self._ZSTD_MAX_CLEVEL}]"
+            )
+        clevel_uint32 = struct.unpack("I", struct.pack("i", clevel))[0]
```
t20100: Need to pass `uint32`. The filter cast this back to `int`

payno: And we are agree that it is cast the same way ?

```python
clevel = -12
clevel_uint32 = struct.unpack("I", struct.pack("i", clevel))[0]
struct.unpack("i", struct.pack("I", clevel_uint32))[0]
>> -12
```
Could we add a comment on this. Because indeed this is a bit tricky and better to highlight that this is done on purpose...

payno (approved):

src/hdf5plugin/_filters.py
t20100: I hesitated to make helper functions to make it more explicit.

payno: To be honest I saw the reverse function a couple of line after but I let the comment.

This is super simple I agree but I would be in favor of helper function (This is easier to test and to associate).

t20100: Added `_to_uint32` and `_from_uint32` helpers in https://github.com/silx-kit/hdf5plugin/pull/395/commits/58afed305cad59c3b13b9b19f23ab16ea7f011f2

Issue #377, opened 2026-02-25
blosc2-grok test fails on Linux


t20100:
[CI fails](https://github.com/silx-kit/hdf5plugin/actions/runs/22392576220/job/64851622302) on blosc2-grok test:
```
  test_blosc2_codec_grok (hdf5plugin.test.TestBlosc2Plugins.test_blosc2_codec_grok)
  Read blosc2 dataset written with blosc2-grok external codec plugin ... python: symbol lookup error: /tmp/tmp.GvdniMRq5B/venv/lib/python3.12/site-packages/blosc2_grok/libblosc2_grok.so: undefined symbol: b2nd_deserialize_meta
```
Maybe related with [latest blosc2-grok version](https://github.com/Blosc/blosc2_grok/releases/tag/v0.3.4)?

attn @FrancescAlted

FrancescAlted:
Interesting.  Any hint @lshaw8317 ?

lshaw8317:
The package fails to find the C-blosc2 shared object - it expects it to be in ``/tmp/tmp.GvdniMRq5B/venv/lib/python3.12/site-packages/blosc2/lib/`` ; could you check that it is there?
 I ran an example on my linux machine and didn't have this problem, so it may be a subtlety with docker.

Also, in order for the bug to raise earlier you could try ``import blosc2_grok`` at the top of the file.

t20100:
Updating the embedded version of c-blosc2 looks to fix the issue, see #378 which should fix the issue.

This is a bit of a special case here, `hdf5plugin` is a Python project but blosc2 is not loaded from Python but through the libhdf5.

lshaw8317:
The solution is quite simple - put `import blosc2_grok` at the top of the file. 

We will work on a more robust version of grok over the next couple of days which will make this import statement unnecessary, but for the moment that should fix things.

t20100:
Indeed it fixes the issue at hand, thanks for looking into it!

I don't plane a release of hdf5plugin in the coming days so this will wait for the improvement in blosc2-grok.

lshaw8317:
Hello,

Ok so this is more difficult than anticipated.
In order for the import statement to be skipped on Linux, one has to run the same line in ``src/CMakelists.txt`` in ``blosc2_grok`` that is used for macOS and Windows
```
target_include_directories(blosc2_grok PUBLIC ${BLOSC2_INCLUDE_DIR})
# link to grok but not to blosc2 (on linux)
[...]

PR #396, opened 2026-06-23
Update CI

t20100:
This PR:
- Update CI to test py314 rather than py313
- Update actions and avoid compression of uploaded artifacts, since they are already compressed
- Fixed missing installed dependencies in CI (to be reworked once #377 is fixed)
- Fixed numpy deprecation warning

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,8 +11,8 @@ jobs:
     name: Build and test source distribution
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v5
-      - uses: actions/setup-python@v5
+      - uses: actions/checkout@v7
+      - uses: actions/setup-python@v6
         with:
           python-version: '3.12'
           cache: 'pip'
@@ -22,17 +22,17 @@ jobs:
       - run: python -m twine check dist/*
       - run: pip install --pre "$(ls dist/hdf5plugin*.tar.gz)[test]"
       - run: python test/test.py
-      - uses: actions/upload-artifact@v4
+      - uses: actions/upload-artifact@v7
         with:
-          name: cibw-sdist
-          path: dist/*.tar.gz
+          path: dist/hdf5plugin-*.tar.gz
+          archive: false
 
   build_doc:
     name: Build documentation
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v5
-      - uses: actions/setup-python@v5
+      - uses: actions/checkout@v7
+      - uses: actions/setup-python@v6
         with:
           python-version: "3.12"
           cache: "pip"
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -108,12 +109,12 @@ jobs:
           CIBW_ARCHS: ${{ matrix.cibw_archs }}
 
           CIBW_TEST_SKIP: "*-*linux_ppc64le"
-          CIBW_BEFORE_TEST: pip install h5py --only-binary ":all:"
+          CIBW_BEFORE_TEST: pip install packaging h5py --only-binary ":all:"
```
t20100, line 112: Tests relies on `packaging`. It's in `[test]` extra requires but there is currently issues with blosc2_grok (#377), so manual install for now until this is fix.

I didn't wanted to add it as a dependency for the project since this is just for on kind of tests.

.github/workflows/release.yml
```diff
@@ -21,18 +21,19 @@ jobs:
...
       - run: pip install --pre "$(ls dist/hdf5plugin*.tar.gz)[test]"
+      - run: pip uninstall -y blosc2_grok  # https://github.com/silx-kit/hdf5plugin/issues/377
       - run: python test/test.py
-      - uses: actions/upload-artifact@v4
+      - uses: actions/upload-artifact@v7
         with:
-          name: cibw-sdist
-          path: dist/*.tar.gz
+          path: dist/hdf5plugin-*.tar.gz
+          archive: false
```
t20100, line 29: do not zip artifacts, they are already compressed

.github/workflows/release.yml
```diff
@@ -137,16 +138,17 @@ jobs:
...
           python-version: ${{ matrix.python-version }}
           cache: 'pip'
-      - uses: actions/download-artifact@v5
+      - uses: actions/download-artifact@v8
         with:
-          pattern: cibw-wheels-*
+          pattern: hdf5plugin-*.whl
           path: dist
           merge-multiple: true
+          skip-decompress: true
```
t20100, line 151: Directly download wheels. There was one wheel per zip file before.
Use `skip-decompress` to avoid unzipping the wheel!

ci/oldest_h5py.txt
```diff
@@ -1,4 +1,6 @@
 numpy < 2 ; python_version <= '3.12'
+h5py == 3.15.0 ; python_version == '3.14'
+h5py == 3.12.1 ; python_version == '3.13'
```
t20100, line 3: Missing oldest h5py for the tests

test/test.py
```diff
@@ -68,8 +68,9 @@ def testBlosc(self):
         self.assertTrue(data.shape[1] == 100, "Incorrect shape")
         self.assertTrue(data.shape[2] == 100, "Incorrect shape")
 
-        target = numpy.arange(numpy.prod(expected_shape), dtype=numpy.float64)
-        target.shape = expected_shape
```
t20100, line 72: numpy depreaction warning

payno (approved):

PR #398, opened 2026-06-23
Doc: add `Example of an installation with conda for test` in contribute

payno:
This PR add a documentation in `contribute` to give some hint on how to set up a conda env for testing / dev.

It has been added to `contribute` section but it can be moved somewhere else.

image


t20100 (approved):
Thanks!

PR #401, opened 2026-06-26
Updated tests

t20100:
This PR:
- enables blosc/blosc2 tests with strings (only supported with hdf5 >= 2.1), closes #364
- remove `import packaging` from the tests: it was an in between situation: it is not a dependency of the project but was needed to run the tests. I propose to avoid using it for now since it is simple rather than adding it as a dependency

### DIFF
--- a/src/hdf5plugin/test.py
+++ b/src/hdf5plugin/test.py
@@ -34,7 +34,6 @@ import unittest
 from typing import Any, cast
 
 import numpy
-from packaging.version import parse as parse_version
 
 import h5py
 import hdf5plugin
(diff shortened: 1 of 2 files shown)


src/hdf5plugin/test.py
```diff
@@ -494,13 +493,17 @@ def _test_strings(
                 self.assertEqual(len(filters), 1)
                 self.assertEqual(filters[0][0], hdf5plugin.FILTERS[filter_name])
 
-    @unittest.skip(reason="segfault (#364)")
+    @unittest.skipIf(
+        h5py.version.hdf5_version_tuple < (2, 1, 0), "hdf5 v2.1.0 required"
+    )
```
payno, line 498: 👍

payno (approved):

PR #403, opened 2026-07-29
Add error checking to zstd HDF5 filter

pimlu:
This PR adds error checking to the zstd HDF5 filter.  Previously it would return uninitialized memory from malloc when any zstd error occurred.

It also prevents proceeding with the result of `malloc(0)` in the abnormal case where `decompSize == 0`.

### DIFF
--- a/lib/hdf5_plugins/ZSTD/src/H5Zzstd.c
+++ b/lib/hdf5_plugins/ZSTD/src/H5Zzstd.c
@@ -66,10 +66,16 @@ H5Z_filter_zstd(unsigned int flags, size_t cd_nelmts, const unsigned int cd_valu
     if (flags & H5Z_FLAG_REVERSE) {
         /* We're decompressing */
         size_t decompSize = ZSTD_getFrameContentSize(*buf, origSize);
+        if (decompSize == 0 || decompSize == ZSTD_CONTENTSIZE_UNKNOWN ||
+            decompSize == ZSTD_CONTENTSIZE_ERROR)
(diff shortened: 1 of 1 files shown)


lib/hdf5_plugins/ZSTD/src/H5Zzstd.c
```diff
@@ -66,10 +66,16 @@ H5Z_filter_zstd(unsigned int flags, size_t cd_nelmts, const unsigned int cd_valu
     if (flags & H5Z_FLAG_REVERSE) {
         /* We're decompressing */
         size_t decompSize = ZSTD_getFrameContentSize(*buf, origSize);
```
pimlu, line 68: Technically `decompSize` should be an `unsigned long long` but on 64-bit systems the outcome is the same.  Let me know if I should update this in the same PR.

t20100:
Thanks for the Pull Request!

This projects embeds copies of third-party hdf5 filters.

For ZSTD, the upstream project is https://github.com/HDFGroup/hdf5_plugins/blob/master/ZSTD/src/H5Zzstd.c.
Please propose your fix to upstream first, we'll update the vendored version of the ZSTD filter once it is fixed upstream.

pimlu:
Oh thanks for pointing me in the right direction! I made a PR in the upstream hdf5_plugins, I will close this PR.

t20100:
Thanks!

For reference, upstream PR: https://github.com/HDFGroup/hdf5_plugins/pull/280

